PR #989, opened 2026-06-29
Show external plugin load failures

giswqs:
## Summary
- retain the latest external plugin load issue by source URL
- expose those issues to Manage Plugins through the existing external plugin load subscription
- show Failed / Failed to load with the loader error instead of an indefinite Loading state when a URL plugin is installed but not registered

## Verification
- npm run lint (passes with existing warnings)
- npm run build -w geolibre-desktop
- browser smoke test: installed live NASA OPERA 0.2.5 from plugins.geolibre.app and verified the row shows Installed, not Loading or Failed

## Notes
The NASA OPERA registry bundle was also hotfixed to 0.2.5 and reduced below 5 MiB. This app-side change prevents future plugin load failures from appearing as a stuck installation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Plugin management now clearly shows when an installed plugin fails to load.
  * Failed plugins now display a visible “Failed” indicator and a translated error message in the details view.
  * Load failure states update automatically after external plugin loading completes.

### DIFF
--- a/apps/geolibre-desktop/src/components/layout/ManagePluginsDialog.tsx
+++ b/apps/geolibre-desktop/src/components/layout/ManagePluginsDialog.tsx
@@ -34,10 +34,12 @@ import {
 } from "react";
 import { useDesktopSettingsStore } from "../../hooks/useDesktopSettings";
 import {
+  getExternalPluginLoadIssues,
   getPluginManager,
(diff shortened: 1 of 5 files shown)


giswqs:
@coderabbitai review

PR #1147, opened 2026-07-09
feat(i18n): complete all 13 locale catalogs + migrate hardcoded panel/dialog strings

giswqs:
## Why

The non-English catalogs had drifted to **~12% coverage** (273 of 2318 keys). Only `common`, `language`, `settings`, and `toolbar` were partially present, so every menu, dialog, and panel outside those fell back to English regardless of the selected language. `en.json` grows with each feature PR, but the locales were last meaningfully updated around #635 — hence "many menus are not translated".

## What

**1. All 13 catalogs to full parity.** Regenerated `de/es/fr/hi/id/it/ja/ko/nl/pt/ru/tr/zh` from `en.json`, going from ~12% → **100%** each. Preserves:
- exact key structure (English key names),
- `{{interpolation}}` placeholders and `<Trans>` markup verbatim,
- per-language CLDR plural forms (Russian `one/few/many/other`; `zh/ja/ko/id` `other`-only; the rest `one/other`).

Terminology follows QGIS-style conventions per language. Brand/format/technical tokens (GeoLibre, MapLibre, DuckDB, GeoJSON, EPSG, SQL, …) left untranslated.

**2. Migrated ~95 hardcoded strings that never reached i18n.** These `title`/`aria-label`/`placeholder`/`label` props stayed English in *every* language because they never went through `t()`:
- `LayerPanel`, `AttributeTable`, `StylePanel`, `RasterSymbologySection`, `AttributeChartDialog`
- `ConversionDialog`, `RasterToolsDialog`, `ProcessingDialog`, `ModelBuilderDialog`

New keys added under `layers.*`, `style.*`, a new `attributeTable.*` section, `rasterSymbology.*`, and `processing.*` (recurring file-picker strings share single `processing.filePicker.*` keys). `en.json`: 2232 → 2319 keys.

**3.** Added `toolbar.command.resetPitch` translations across all locales for #1146.

## Validation

- `tests/i18n-catalogs.test.ts`, `tests/i18n-languages.test.ts`, `tests/add-data-i18n.test.ts` — **49/49 pass**; per-locale coverage reports **13 × 100%**.
- No extra/misspelled keys; all interpolation placeholders preserved (enforced by the catalog test).
- `tsc -b` clean for all touched files (the only errors are pre-existing `cesium` module-resolution in `packages/map`, unrelated to this PR).

The seeded translations are machine-assisted and open to native-speaker review per `docs/i18n.md`.

## Summary by CodeRabbit

* **New Features**
  * Added broader multilingual support across attribute, style, layer, processing, and raster-related panels.
  * Expanded localized text for chart controls, file pickers, search tools, refresh settings, and layer/style dialogs.

* **Bug Fixes**
  * Replaced remaining hardcoded labels, tooltips, placeholders, and accessibility text with translated strings.
  * Improved fallback messaging for errors, invalid values, and empty states to match the selected language.


giswqs:
Addressed the automated review in `24cb0fa`:

**Visible-text gap (fixed).** The review correctly caught that several `title`/`aria-label` attrs were migrated to `t()` while the *adjacent visible text* stayed hardcoded English — so screen readers were localized but sighted non-English users still saw English. Now migrated:
- `AttributeTable`: all 10 toolbar button labels (Edit/Save/Add field/Calculate/Fields/Explore/Statistics/Charts/Dashboard/Export) + the Fields dropdown (Show fields / No fields / Show all fields).
- `LayerPanel`: the "Layers" heading (reuses `sharedRail.layers`).
- `StylePanel`: the "Style" heading and "Style - {{name}}" headings (4 sites).

Reused existing keys where the English matched exactly; added 11 new keys and translated them into all 13 locales using each catalog's existing terminology. All 13 stay at 100% parity; placeholder/key-parity tests green; `tsc` clean for touched files.

**Bundle size (follow-up, not in this PR).** The perf note is valid: `i18n/index.ts` eager-loads all 14 catalogs, and taking coverage 12%→100% grows that payload. I've left it out of scope here because (a) `i18n/index.ts` isn't touched by this PR and the eager-load predates it, and (b) switching to lazy per-locale loading changes the init/`changeLanguage` flow and deserves its own PR with an e2e check on language switching + the `?locale` embed path. Happy to open that as a fast-follow.

giswqs:
Addressed the re-review in `6bcc10e`:

- **`ConversionDialog` dead placeholders (fixed).** Good catch — the lon/lat/layer-name inputs are seeded with defaults via `useState("longitude"/"latitude"/"data")`, so the placeholder never rendered and the visible pre-filled text stayed English. Rather than translate the defaults (they're expected CSV column names / a layer id — data values that should match the user's headers, per `docs/i18n.md`'s "don't translate field-name/format tokens"), I reverted the three placeholders to literals and dropped the now-orphaned `processing.conversion.*` keys from `en.json` + all 13 locales.

Two minor low-confidence points I intentionally left as-is:
- **Mixed key reuse for button visible text** (`attributeTable.charts`/`dashboard` reused for both `aria-label` and the visible `<span>`, vs new `buttons.*` keys elsewhere): intentional — where an existing key's English value is exactly the button label, reusing it avoids a duplicate string to translate; new `buttons.*` keys were only added where no exact match existed.
- **Output-filename fallback `?? "output"`**: same category as the column-name defaults — a suggested-value/data token, not UI chrome, so left untranslated.

giswqs:
`e4e5cb2`: completed a full i18n sweep of the three repeatedly-flagged panels (StylePanel, AttributeTable, LayerPanel) rather than fixing leftovers piecemeal. This migrates all remaining visible hardcoded text — symbology/extrusion/raster/mode controls, column-header menus, the field-calculator and add/delete-field dialogs, layer-action menus, and the auto-refresh/metadata/remove dialogs — ~192 new `t()` sites and 161 new keys (reusing existing keys on exact matches), all translated across the 13 locales at 100% parity. Includes the two explicitly-flagged spots (`Layer order (default)` option, `Attribute table` heading). Format names, data-value defaults, and expression code stay literal per `docs/i18n.md`. Local: frontend suite 2228 pass, catalogs 13×100%, tsc clean.

PR #1273, opened 2026-07-16
fix: use CSS :lang() pseudo-class for CJK vertical text rotation

Craun718:
Hello! 

I found that a "rotate" is used in the sidebar. I guess that is to ensure display English to outdise in vertical display to conform read habits. But it will make the CJK language display upside down. 

## befor：
image
image

## after：
image
image

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved vertical sidebar labels for Chinese, Japanese, and Korean content by keeping characters upright on larger screens.


giswqs:

Issue #952, opened 2026-06-27
[Bug]: Hide the Plugin panels on project load


spatialthoughts:
### Before you start

- [x] I searched the [existing issues](https://github.com/opengeos/GeoLibre/issues) and did not find a duplicate.
- [x] This is a bug report, not a usage question (questions belong in [Discussions](https://github.com/opengeos/GeoLibre/discussions)).

### What happened?

When a user opens a project via the gallery link, the plugin panels are open and blocking the view. 

<img width="776" height="680" alt="Image" src="https://github.com/user-attachments/assets/e932a166-3fcd-47e0-8cce-c9ecbd353fe2" />

### Steps to reproduce

Visit https://web.geolibre.app/?url=https://share.geolibre.app/spatialthoughts/maritime-piracy.geolibre.json

### Expected behavior

Basemap panel should be hidden

### How are you running GeoLibre?

Web

### Operating system and browser

MacOS

renevandervelde:
Hi Spatialthoughts, 

As a beta tester, the developer of GeoLibre, Qiusheng Wu, granted me co-admin rights in Github to reopen and re-assign issues in GitHub. Normally, I would only use this if an issue needs extra attention, extra work, or to schedule additional beta testing.

In this case, I made temporary use of the possibility to reopen an issue, but it will only be for the duration of 48 hours (and I will close this topic manually next Monday). 

Be assured: currently no extra work or beta testing is  necessary for the issue you raised in this GitHub topic. The additional 'wont fix' label I just assigned to your topic is to indicate that the programming and enhancement has already been done, and this topic is currently just reopend by me for another purpose.

The reason I reopened your topic on GitHub **to compliment you, ensuring that you, and potentially others, can see this compliment publicly.**

**Stunning map**
What a stunning Story Map you have created. I can easily imagine this being a perfect fit for a showcase like "perfect examples of beautiful story maps made with GeoLibre".

I do not think such a showcase GeoLibre library exists yet, but I can certainly see the GeoLibre developer Qiusheng Wu asking for your permission to include this project if he ever looks for inspiring examples to motivate other users by presenting maps like yours in such an example library. 

Of course, it is also completely understandable if the data displayed in this Story Map is sensitive and cannot be shared publicly.

Regardless, I just wanted to compliment you on this beautiful piece of work!

Kind regards,

René (a fellow GeoLibre user)


[...]

PR #1346, opened 2026-07-19
fix(time-slider): restore the dock open, add mosaic NoData, fix the WASM nodata border

giswqs:
## Problem

Reported against https://web.geolibre.app/?url=https://share.geolibre.app/giswqs/pace-chla-3.geolibre.json — a project saved with the Time Slider active and its dock expanded reopens with the **dock hidden**, leaving only the small clock toggle at bottom-left.

The saved project is correct (`activePluginIds` contains the plugin, `settings` carries `collapsed: false` and the source), and the plugin does activate. The dock is created and added — then hidden.

## Cause

`PluginManager.restoreProjectState` wraps the scoped app so every control added during a restore is collapsed, so a loaded project does not bury the map under expanded panels (#952):

```ts
const scopeForRestore = (id: string): GeoLibreAppAPI =>
  scopeAppToPlugin(app, id, { onControlAdded: collapseRestoredPanel });
```

`TimeSliderControl` exposes the duck-typed `collapse()` this looks for, and its collapsed style is `display: none` — so the dock is hidden outright, overriding the project's `collapsed: false`.

That heuristic is right for panels which auto-expand on activation and carry no saved state. It is wrong for a plugin whose collapsed state *is* part of the project.

It also corrupted the project on the way through: `collapse()` mutates the control's live state, so `getProjectState` then reported `collapsed: true` and any re-save wrote that back into the file — each open worse than the last.

## Fix

Adds `restoresPanelCollapseState` to `GeoLibrePlugin`. A plugin that declares it is exempt from the restore-time collapse sweep, because the restored config already decides whether its panel opens. Declared on the Time Slider only, so #952 behaviour is unchanged for every other plugin.

## Verification

Driven in a real browser against the reported share link: the dock restores `attached=true visible=true ts-collapsed=false`, with 0 console errors.

Tests: a new case asserting an opted-out plugin stays expanded, alongside the existing #952 cases which still pass (31 in the file, 3357 in the suite).

## Related upstream

Two other issues from the same report are fixed upstream and will land here as dependency bumps once released:

- opengeos/maplibre-gl-time-slider#25 — mosaic sources had no NoData option
- opengeos/cog-tiler-wasm#23 — the WASM engine drew a blue border around nodata boundaries (bilinear interpolation blended the `-9999` sentinel into edge pixels, which then clamped to the bottom of the `jet` ramp)

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored projects now preserve panel expanded/collapsed state when plugins persist their own panel state.
  * The time slider panel no longer unexpectedly collapses after project restoration.
* **Tests**
  * Added a case to confirm panel auto-expand behavior during project restore when a plugin opts into restoring its own collapse state.
* **Chores**
  * Updated time slider dependency to the latest patch/minor version.


giswqs:
Updated: the two upstream fixes are released and this PR now bumps to them.

| Package | Version | Fixes |
|---|---|---|
| [`cog-tiler-wasm@0.3.1`](https://github.com/opengeos/cog-tiler-wasm/pull/23) | 0.3.0 → 0.3.1 | WASM engine drew a blue border around nodata boundaries |
| [`maplibre-gl-time-slider@1.4.0`](https://github.com/opengeos/maplibre-gl-time-slider/pull/25) | 1.3.0 → 1.4.0 | mosaic sources had no NoData option |

So all three reported issues are covered here.

### Verified end to end

Driven in a browser against the reported project with the **published** packages installed (not local builds): dock restores attached and visible, 0 console errors, 0 failed requests, chlorophyll layer and colorbar render.

The border fix was measured separately against real PACE L2 COGs (`float32`, `NoData=-9999`, `jet` over `[0,30]`), 15 tiles at z6:

| | opaque px | pixels on jet's exact floor | warm end (real high-chla) |
|---|---|---|---|
| before | 24,002 | **13,906 (57.9%)** | 111 (0.46%) |
| after | 16,636 | **0 (0.00%)** | 237 (1.42%) |

Gates: `npm run test:frontend` 3357 pass / 0 fail, `npm run build` clean.

### Note for anyone re-testing

The original share link (`share.geolibre.app/giswqs/pace-chla-3.geolibre.json`) started returning 404 partway through this work, so the final verification serves a saved copy of that project locally. The project content is unchanged.

PR #1357, opened 2026-07-20
fix(whitebox): hide WASM-absent catalog tools in local mode (#1355)

### LINKED ISSUE #1355: [Bug]: Missing input paramters on Whitebox UI
### Before you start

### What happened?

Image

### Steps to reproduce

1. open: https://web.geolibre.app/
2. open whitebox toolbox
3. open assign projection raster/vector/lidar

### Expected behavior

that there would be file names to select and projections to assign. there are no input/output selections

https://web.geolibre.app/

_No response_

### Sample data

_No response_

### Screenshots or logs

_No response_

giswqs:
## Summary

In the Whitebox toolbox's local (Run locally / WASM) mode, opening **Assign Projection Vector / Raster / Lidar** showed only "This tool has no parameters" with no input or projection fields (#1355).

**Root cause:** these three tools ship in the Whitebox catalog snapshot with an empty parameter list, and they are **not present in the `geolibre-wasm` binary** at all. The local-mode merge kept catalog tools even when the binary did not implement them, so they rendered a dead-end form and, if Run, failed with `tool not found`. Six catalog tools total are absent from the WASM binary (`assign_projection_{vector,raster,lidar}`, `reproject_lidar`, `false_colour_composite`, `true_colour_composite`); the three `assign_projection_*` ones have no params, which is the reported dead-end.

## Fix

When the WASM manifests load, the binary's tool set is the ground truth for what can run locally, so `mergeWasmToolManifests` now drops catalog tools the binary lacks, exactly as locked "pro" tools that cannot run are already hidden. When manifest enumeration fails (an empty manifest list), the whole catalog is kept as a display-only fallback so the toolbox is not emptied.

This is a GeoLibre-only change; no upstream package or release was needed. Only the local-mode merge path is affected (sidecar mode, where these tools do run and report live parameters, is untouched).

## Verification

Drove the real web build with Playwright in both light and dark themes:

- Searching "assign" no longer lists the three `assign_projection_*` dead-end tools.
- The "Projection and Georeferencing" list keeps the WASM-runnable tools (Reproject Raster/Vector, Georeference Raster From Control Points) and no longer shows the WASM-absent Reproject Lidar.
- Tool count drops from 755 to 749 (exactly the six WASM-absent tools), and runnable tools still render their parameters.

Also added a `mergeWasmToolManifests` unit test for the drop, clarified the existing empty-manifest fallback test, and ran the full frontend suite (green).

Fixes #1355

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved WASM tool availability handling.
  * Retained catalog tools as a display-only fallback when manifest loading returns no tools.
  * Removed catalog tools that are not implemented by a successfully loaded WASM binary.
  * Preserved tool parameter reconciliation and source information during merging.

### DIFF
--- a/packages/processing/src/wasm-client.ts
+++ b/packages/processing/src/wasm-client.ts
@@ -188,9 +188,20 @@ export async function listWasmToolManifests(): Promise<WhiteboxTool[]> {
  * metadata but takes the manifest's parameters; the GeoLibre-authored tools that
  * never appear in the catalog are appended.
  *
+ * When the WASM manifests loaded, the binary's tool set is the ground truth for
+ * what can actually run locally, so a catalog tool the binary does **not**
+ * implement is dropped: it cannot run in WASM mode and would otherwise render as
+ * a dead-end form (the `assign_projection_*` tools carry no catalog params, so
+ * the dialog just says "This tool has no parameters", GeoLibre#1355) or, for
+ * ones the catalog does parameterize (`reproject_lidar`, the colour composites),
+ * fail on Run with "tool not found". This mirrors how locked "pro" tools that
+ * cannot run are hidden entirely. When manifest enumeration failed (an empty
+ * `wasmTools`), the whole catalog is kept as a display-only fallback rather than
+ * emptying the toolbox.
+ *
  * @param catalogTools - Tools from the Whitebox catalog snapshot.
  * @param wasmTools - Every WASM tool manifest ({@link listWasmToolManifests}).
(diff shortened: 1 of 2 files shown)


giswqs:
Superseded. Rather than hiding these tools, the root cause is being fixed upstream: `assign_projection_{raster,vector,lidar}` are now implemented in geolibre-rust (opengeos/geolibre-rust#43) so they run in the WASM build with proper parameters. A new GeoLibre PR will bump `geolibre-wasm` to consume the release.

PR #1510, opened 2026-07-28
chore(deps): bump maplibre-gl to 6.3.0

### LINKED ISSUE #1489: maplibre-gl v6 blocked: plugin ecosystem still ships v5-era builds
`maplibre-gl` v6 cannot land until the plugin ecosystem ships v6-compatible builds. Tracking what has to happen first, so the next Dependabot bump (#1482) is not re-attempted blind.

## Why it is blocked

MapLibre v6 is an **ESM-only** distribution with **no default export** and **no CJS build** (the UMD and `maplibre-gl-csp` bundles are gone). See the [v5 → v6 migration guide](https://maplibre.org/maplibre-gl-js/docs/guides/v5-to-v6-migration-guide/).

Three separate breakages result, and only the third is ours to fix:

### 1. Third-party packages ship dist compiled against v5's default export

`import maplibregl from "maplibre-gl"` is a hard bundling error against v6 (`"default" is not exported by maplibre-gl.mjs`). These 11 installed packages still do it in their published `dist/`:

**opengeos-owned (9)** — need a namespace-import migration + rebuild + release:

- `maplibre-gl-3d-tiles`
- `maplibre-gl-components`
- `maplibre-gl-duckdb`
- `maplibre-gl-geo-editor`
- `maplibre-gl-geoagent`
- `maplibre-gl-overture-maps`
- `maplibre-gl-raster`
- `maplibre-gl-streetview`
- `maplibre-gl-swipe`

**external (2)** — we can only wait or shim:

- `@esri/maplibre-arcgis`
- `@geoman-io/maplibre-geoman-free`

Note `maplibre-gl-raster` uses the mixed form `import maplibregl, { Popup } from "maplibre-gl"`, so a scan that only looks for the bare default import will miss it.

Re-check with:

```bash
grep -rlE 'import +[A-Za-z_$][A-Za-z0-9_$]* *(,[^;]*)? *from *["'"'"']maplibre-gl["'"'"']|require\( *["'"'"']maplibre-gl["'"'"'] *\)' \
  --include='*.js' --include='*.mjs' --include='*.cjs' \
  node_modules/maplibre-gl-* node_modules/@esri/maplibre-arcgis node_modules/@geoman-io/*
```

### 2. v6 dropped CJS, which breaks the Node test suite

Those packages are dual-published, and their `.cjs` entries `require("maplibre-gl")`. v6's `exports` map has only `types` and `import` conditions, so CJS resolution fails outright:

```
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined in node_modules/maplibre-gl/package.json
```

`npm run test:frontend` fails **29 tests** on this. A Vite-level shim cannot help — `node --test` does not go through Vite.

### 3. GeoLibre patches the MapLibre namespace (ours to fix)

`packages/map/src/globe-popup-occlusion.ts` installs globe-occlusion behavior by assigning over the namespace:

```ts
api.Popup = GeoLibrePopup;   // map-controller.ts calls installGlobePopupOcclusion(maplibregl)
```

Under v5's mutable CJS/UMD default export this worked. A v6 ESM namespace object is getter-only, so it throws at boot:

```
TypeError: Cannot set property Popup of #<Object> which has only a getter
```

The app never mounts the map canvas — **22 of 23 e2e tests fail**. This needs a different mechanism (the patch exists so third-party plugins constructing `new maplibregl.Popup()` also get occlusion, so simply exporting a `GeoLibrePopup` for our own call sites is a behavior change worth deciding on deliberately).

## What is already done

A local `wip` branch carries the GeoLibre-side migration, verified with `tsc -b` clean, `eslint` 0 errors, and a successful `vite build`:

- **Dedupe to a single copy.** The Dependabot lockfile resolved *both* 5.24.0 (hoisted, to satisfy `@maplibre/maplibre-gl-directions`' `^5.0.0` peer — there is no v6-compatible release of it) and 6.0.0 (nested in the three workspaces that declare it), so `Map` was a different nominal type across package boundaries. That is what the original CI errors (`Argument of type 'SwipeControl' is not assignable to parameter of type 'IControl'`) actually were — not an API change. Fixed with a root `overrides` entry.
- **33 files** migrated to `import * as maplibregl from "maplibre-gl"`.
- **v6's tightened typing**: `set/getPaintProperty` and `set/getLayoutProperty` are now generic over `keyof AllPaintProperties` / `keyof AllLayoutProperties`, and `Map#on`/`off`'s catch-all overload narrowed from `type: string` to `type: keyof MapEventType` (which breaks binding third-party events such as Geoman's `gm:*`). Added `packages/map/src/dynamic-style-property.ts` so the unavoidable casts for runtime-computed property names live in one documented place.

## Suggested order

1. Migrate + release the 9 opengeos packages (namespace imports, rebuild, verify the `.cjs` entry no longer `require`s maplibre-gl).
2. Rework `globe-popup-occlusion.ts` off namespace patching.
3. Wait on / track `@esri/maplibre-arcgis` and `@geoman-io/maplibre-geoman-free`. `@maplibre/maplibre-gl-directions` also still pins `maplibre-gl@^5.0.0` as a peer.
4. Re-run the bump and re-check the frontend + e2e suites, not just the build — the build passes with a shim while the app is completely broken at runtime, so a green build alone proves nothing here.

Until then, `maplibre-gl` should stay on 5.x and Dependabot bumps to v6 should be closed with a pointer to this issue.

giswqs:
Ready for review. Everything in #1489 is now done here and verified: both external blockers shipped v6-compatible builds, so the shim that was the open policy question is **gone rather than kept**.

Depends on #1509 (blocker 3), already merged.

### Blocker 2 — the frontend suite ✅

Not the plugin packages: their `exports` maps all publish a correct `import` condition. It was **our own module scope**. Every workspace package is `"type": "module"`, but the root `package.json` is not, so tsx compiled all 256 test files to **CJS** — which selects the `require` condition and loads the `.cjs` entries that `require("maplibre-gl")`.

`tests/package.json` with `{"type": "module"}` puts them in ESM scope. The 29 `ERR_PACKAGE_PATH_NOT_EXPORTED` failures go to **zero**. Measured both ways:

| | tests CJS | tests ESM |
|---|---|---|
| **5.24.0** | 4234 pass / 0 fail | 31 fail — v5 ships no ESM build at all |
| **6.0.0** | 29 fail, all `ERR_PACKAGE_PATH_NOT_EXPORTED` | 0 of that class |

It is a swap, not an addition, which is why it ships here and could not land ahead of the bump.

### Blocker 1 — the two external packages ✅ resolved upstream, shim deleted

`@esri/maplibre-arcgis` and `@geoman-io/maplibre-geoman-free` both default-imported `maplibre-gl` in their published ESM, which v6 does not provide. An earlier revision of this PR carried a ~40-line self-deleting shim for that. Both have since shipped:

- **`@esri/maplibre-arcgis` 1.3.1** — imports `{ AttributionControl }` by name; peers `~5.24.0 || ~6.3.0`.
- **`@geoman-io/maplibre-geoman-free` 0.9.0** — Geoman's MapLibre v6 release. Peers `maplibre-gl >=6.0.0 <7.0.0`, ESM-only (the UMD and CJS entries are gone, since neither could resolve v6), and its bundle now opens with `import * as e from "maplibre-gl"`.

So the shim comes out in all three places a module could enter the tree — the Vite plugin, the Node loader hook for `node --test`, and the `optimizeDeps.exclude` entry that kept the dev server's dependency optimizer from choking before the plugin was consulted — along with `tests/maplibre-shim-parity.test.ts`, which existed to keep the two lists in step. No third-party code is rewritten at build time any more.

Geoman 0.9.0 needs no code change on our side: `Geoman`, `defaultLayerStyles`, `waitForGeomanLoaded()`, `destroy({ removeSources })`, `disableAllModes()` and `loaded` all keep their shapes. Its other v6 requirement — set the worker URL before the first `Map` — is already satisfied by the fix below.

## The bug none of the gates would have caught

v6 ships its worker as a **separate file**, located at runtime with `new URL("./maplibre-gl-worker.mjs", import.meta.url)`. That is a computed string, not a static literal, so no bundler can see it: the asset was never emitted, and at runtime the URL resolved next to the hashed app chunk (`/assets/maplibre-gl-worker.mjs`) where nothing exists. In the web build the SPA fallback answers that request with `index.html`, so it does not even 404 — the worker is handed HTML and the request **hangs**.

The map still drew, so build, typecheck, and most of e2e were all green. Tile parsing was quietly degraded: in the same 25-second window, **39 tile requests before the fix, 68 after**. It surfaced only because `pwa.spec.ts` waits for `networkidle`, which that hanging request never let happen.

Fixed by importing the worker through Vite (`?worker&url`, so its `./maplibre-gl-shared.mjs` import is bundled too) and pointing `setWorkerUrl` at the emitted asset.

## The second bug none of the gates would have caught

v6 also split `Camera` out of `Map`: the transform moved to `map._camera.transform` and `Map` has none of its own. This branch had already migrated GeoLibre's own readers (`globe-popup-occlusion.ts`, `readCameraAltitude`), but not the third-party code we hand the map object to.

`@deck.gl/mapbox`'s `getViewport()` reads `map.transform.height` on **every** deck render, so with `transform` gone it threw `Cannot read properties of undefined (reading 'height')` once per frame and no deck.gl overlay drew anything: the maplibre-gl-raster GPU engine, point clouds, 3D Tiles, the deck.gl-viz layers. Reproduced with Add Data > Raster Layer, the GPU engine, and the NLCD land-cover sample: on `main` the raster renders with a clean console; on this branch nothing rendered. deck.gl still reads `map.transform` on 9.3.10, the current release, so there is no version to upgrade into; `maplibre-gl-components` (ControlGrid) and `maplibre-gl-splat` (GaussianSplatControl) read it too.

Nothing caught it because a bare map boot never constructs a deck overlay: build, typecheck, lint, the frontend suite and e2e were all green while every deck.gl layer was broken.

`packages/map/src/map-transform-compat.ts` puts the getter back on `Map.prototype`, delegating to the camera's transform, installed at `map-controller.ts` module load so it is in place before any `Map` exists. It applies only when `Map` really has no `transform`, so it is inert on v5 and self-disabling if MapLibre brings it back. `_nearZ`/`_farZ` are aliased onto the transform as well, since v6 renamed them to the public `nearZ`/`farZ` and deck.gl would otherwise quietly fall back to its own near/far multipliers and lose depth precision on pitched and terrain views.

## Everything else

- **30 files** migrated to `import * as maplibregl from "maplibre-gl"`.
- **Root `overrides` pin** — `@maplibre/maplibre-gl-directions` still peers `maplibre-gl@^5.0.0`, which otherwise resolves a second hoisted copy and makes `Map` a different nominal type across package boundaries (the real cause of the original `SwipeControl is not assignable to IControl` errors). One copy installs now.
- **`packages/map/src/dynamic-style-property.ts`** — v6 made `set/getPaintProperty` and `set/getLayoutProperty` generic over `keyof AllPaintProperties`. Property names we compute at runtime (`Object.entries(spec.paint)`, `` `${prop}-transition` ``) need a cast; they live in one documented module instead of scattered call sites.
- **Geoman's `gm:*` events** — `Map#on`/`off`'s catch-all overload narrowed from `type: string` to `keyof MapEventType`.
- **`GeoAgentSyncableTools`** — its map slice now uses method syntax, so a real `Map` still satisfies it under `strictFunctionTypes` while test fakes still do.
- **GeolocateControl factory** — `map-controller.test.ts` swapped `maplibregl.GeolocateControl` to inject a fake, which the sealed v6 namespace rejects (same root cause as #1509). Construction now goes through a factory the test can override.
- **jsPDF imported by name** — its `node` export condition resolves a CJS bundle whose *default* is the module object, not the constructor. Only surfaced once the tests moved to ESM scope; the browser build was always fine.
- **`scripts/patch-maplibre-read-buffer.mjs` removed** (from #1874, and the reason the merge with `main` did not just resolve textually). It rewrote `.STREAM_READ` to `.STREAM_DRAW` in `node_modules/maplibre-gl/dist/maplibre-gl.js` from a workspace `postinstall`, and v6 ships no such file — the hook exits 1 with ENOENT and npm rolls the whole install back. The patch is obsolete rather than merely relocated: it targeted the fullscreen-triangle pixel-pack readback, and v6 removed that path entirely (v5's bundle has 8 `PIXEL_PACK_BUFFER` and 6 `_fullscreenTriangle` references; v6's dist has zero of each, across every `.mjs` it ships). With no READ-usage buffer left there is no Chromium readback-shadow warning to silence, so the script, its `postinstall`, its Dockerfile COPY and its test all come out together.

## Verification

- `test:frontend` **5945 passed / 0 failed** (1 skipped) — including `geo-editor-plugin.test.ts`, which statically imports Geoman and was the single test the Node loader hook existed for. Coverage above the floors on all three axes.
- `tsc -b && vite build` clean · `lint` 0 errors · `test:worker` clean. Geoman still lands in its own `maplibre-geoman` chunk.
- **Real browser** (Chromium on the GPU, production build via `vite preview`): map boots, globe renders, zero console errors. The NLCD land-cover COG renders through the deck.gl-backed `maplibre-gl-raster (GPU)` engine with a clean console and no diagnostics, which is the check that catches the transform regression above. Activated the GeoEditor plugin, drew a polygon on the globe, and it rendered and landed in a Sketches layer — Geoman's full toolbar renders (Polygon, Line, Rectangle, Circle, Marker, Freehand, Text, Drag, Edit, Rotate, Cut, Delete, Union, Difference, Simplify, snapping, topology, undo/redo). Still zero console errors through the draw.
- `test:e2e` 19/23 locally on the earlier revision. All four failures (`a11y`, both `export` specs, `layer-panel`) reproduced on that machine independent of this branch and each passed in isolation — they look like local resource contention, and CI is the tiebreaker.

Not run locally: backend/rust suites, which this cannot affect.

## Summary by CodeRabbit

* **New Features**
  * Upgraded mapping capabilities to MapLibre GL v6.
  * Added improved support for dynamic map styling and runtime layer updates.
  * Added more reliable map worker loading for desktop maps.

* **Bug Fixes**
  * Improved compatibility for camera movements, geolocation controls, custom map protocols, and mapping tools.
  * Preserved viewport history correctly during scripted story and flight animations.
  * Improved compatibility with updated mapping and PDF export functionality.


giswqs:
Merged `main` and moved the bump to **6.3.0** (6.2.0 was the ask; 6.3.0 shipped in the meantime and `^6.2.0` resolves straight to it, so this takes one bump instead of two). The write-up above still describes the original 6.0.0 landing — everything in it holds; below is only what changed since.

### Merge with `main`

Four conflicts. Three were files `main` touched in the same import lines this branch migrated off the default `maplibre-gl` import; the fourth was the root `overrides` block, where both sides added an entry.

`useCommandBridge.ts` conflicted as **binary**: this branch's side still carried the literal NUL bytes in `.join("\0")` that `main` has since replaced with the escape sequence, so git would not merge it as text. Taken from `main` with the namespace import re-applied.

### Code from `main` that needed the same v6 treatment

- `NetcdfSampleMarkers`, `useNetcdfIdentify`, `useCogSpectralIdentify` — default import → namespace import.
- `useCollaboration`, `useViewportHistory` — their `storyCameraToken` / `flightCameraToken` `moveend` handlers typed the parameter as a standalone shape, which v6's listener type rejects (weak-type check: no properties in common with the real event). They now intersect with `MapLibreEvent`, the way `map-controller.ts` already does. The tokens ride along as `eventData` on a real event, so this is also the truer type.
- `MapController.readCameraAltitude` (#1816) read `map.transform`, which v6 moved to `map._camera` when `Map` stopped extending `Camera`. It probes both now, matching `globe-popup-occlusion.ts`.

### One more ESM-scope fallout

`workers/collab` gains `"type": "module"`. `tests/collab-comment-validate.test.ts` landed on `main` *after* this branch put `tests/` in ESM scope, and it imports through `workers/collab/src/comment-validate.ts` — a bare `export * from "@geolibre/collab-core/comment-validate"`. In CJS scope that re-export is invisible to Node's named-export detection, so all ten names failed to resolve. The worker is already ESM in every other sense (`export default {}`, and wrangler bundles it as ESM); `workers/collab-node` is declared this way already.

### 6.1.0 → 6.3.0 itself

No source change. The separate worker file is still `dist/maplibre-gl-worker.mjs`, so the `?worker&url` + `setWorkerUrl` wiring still emits and resolves the asset (verified in `dist/assets/`). Both shimmed packages still need the shim — the Vite plugin errors the build when a listed package stops matching, and the build is green, so neither `@esri/maplibre-arcgis` nor `@geoman-io/maplibre-geoman-free` has published a v6-compatible ESM build yet.

### Verification

- `test:frontend:coverage` — **5818 passed / 0 failed / 1 skipped**, 86.30 lines · 84.41 branches · 73.39 functions, all above the floors.
- `tsc -b && vite build` clean · `lint` 0 errors (40 pre-existing warnings) · `test:worker` clean · `workers/collab` typecheck clean.
- `pre-commit run --files <changed>` clean, including the `npm build` hook.

Not run locally: e2e, backend, and rust suites — leaving those to CI.

giswqs:
`@esri/maplibre-arcgis` **1.3.1** — and with it, **half the shim is gone.**

1.3.1 peers `maplibre-gl: ~5.24.0 || ~6.3.0`, and its published ESM now imports by name (`import{AttributionControl as Zt}from"maplibre-gl"`). The only default imports left in the package are in its README and `examples/`, outside the module graph.

The guard did its job. With 1.3.1 installed and the entry still in the list, the build failed with exactly the message it was written to produce:

```
RolldownError: maplibre default-import shim matched nothing in: @esri/maplibre-arcgis.
If the package now ships a v6-compatible build, remove it from SHIMMED_PACKAGES in
vite-plugins/maplibre-default-import-shim.ts.
```

Removed from all three entry points — the Vite plugin's `SHIMMED_PACKAGES`, the Node loader hook, and `optimizeDeps.exclude`, so the dev server pre-bundles it normally again. `@geoman-io/maplibre-geoman-free` still needs the shim, so the machinery and the parity test stay. The parity test keeps its minified-form assertion (`import Zt from"maplibre-gl"`) even though it was this package that published that shape — any bundle may be minified without the space, and the regex still has to handle it.

### Verified in the browser, on the dev server

`optimizeDeps.exclude` only matters on the dev server, so that is where this was checked rather than against the production build.

- Vite now serves the package's maplibre-gl import as `import { b as Ki, jt as xm, mt as up } from "/node_modules/.vite/deps/maplibre-gl-*.js"` — named, off the optimized dep, no rewrite involved.
- All **8 exports** resolve: `AttributionControl, BasemapSession, BasemapStyle, EsriAttribution, FeatureLayer, HostedLayer, VERSION, VectorTileLayer`.
- **Add Data → ArcGIS Layer** against a live public FeatureServer (`USA_Major_Cities_`) adds the layer and renders it — layer row in the panel, points drawn across the country, **0 console errors**.

`test:frontend:coverage` 5818 passed / 0 failed (86.30 / 84.41 / 73.39) · build, lint, `test:worker` clean · `pre-commit` clean including the `npm build` hook.

**One thing to watch:** the new peer is `~6.3.0`, i.e. pinned to the 6.3 minor. Our range is `^6.3.0`, so whenever maplibre-gl ships 6.4.0 this will resolve into a peer conflict until Esri widens it. Worth expecting on the next Dependabot maplibre PR.

PR #1557, opened 2026-07-29
feat(plugins): expose queryZarrLayer for click-to-value on native Zarr layers

### LINKED ISSUE #1555: Expose zarr-layer queryData (or Identify) for addZarrLayer layers — click-to-value on native Zarr layers
### Summary

Expose `@carbonplan/zarr-layer`'s `queryData` on layers added through `addZarrLayer`, so a plugin can implement click-to-value / Identify (and region statistics) on a natively-rendered Zarr layer.

### Background

`zarr-layer` v0.6.1 ships exactly the primitive needed for pixel inspection, on the `ZarrLayer` instance:

```ts
queryData(geometry: QueryGeometry, selector?: Selector, options?: QueryOptions): Promise<QueryResult>
```

- `geometry` is GeoJSON — a `Point` for click-to-value, or `Polygon`/`MultiPolygon` for region stats.
- It handles the CRS reprojection and fill-value masking internally: you pass a WGS84 `[lng, lat]` (straight from a map click) and it reads the value in the store's native grid.
- Returns `{ [variable]: values, dimensions, coordinates }`, with coordinates in the store's source CRS.

### The gap

`addZarrLayer(name, url, options)` returns only a **layer id**, and the `ZarrLayer` instance is module-private, so a plugin has no way to call `queryData` on it. There is no equivalent in `GeoLibreAppAPI` today (`queryData` isn't referenced anywhere in the app).

This already has a precedent: **`setZarrLayerSelector(layerId, selector)`** reaches the same live instance by id to re-select dimensions. A query passthrough is the read-side counterpart of that method.

### Use case

The [Open Climate Service plugin](https://github.com/dhis2/open-climate-service-geolibre-plugin) renders published climate datasets via `addZarrLayer` and wants **click-to-value Identify** (value at the clicked point, for the current time selection). The renderer already has the data and the CRS math via `queryData`; the plugin just can't reach it on the native path. (Reimplementing it plugin-side with a separate `zarrita` point-read duplicates the reprojection + chunk-indexing the renderer already does.)

### Proposal (either direction resolves it)

1. **A passthrough on `GeoLibreAppAPI`**, mirroring `setZarrLayerSelector`:

   ```ts
   queryZarrLayer?: (
     layerId: string,
     geometry: QueryGeometry,
     selector?: Record<string, number | string>,
     options?: { includeSpatialCoordinates?: boolean; signal?: AbortSignal },
   ) => Promise<QueryResult | null>;   // null when there is no live layer with that id
   ```

   A thin delegate to the instance's `queryData`, resolving `null` for an unknown id (same contract as `setZarrLayerSelector` resolving `false`).

2. **Or wire it into GeoLibre's own Identify** for `addZarrLayer` layers, so click-to-value works without any plugin code — consistent with the pixel inspection other opengeos plugins (NASA OPERA, HyperCoast) already offer.

Option 1 is the smaller, more general change and lets each plugin present the value in its own UI; option 2 is more turnkey but bakes in the presentation.

### Benefit

`queryData` is already the right primitive and is currently unreachable for the native path — exposing it (or an Identify built on it) benefits every plugin that renders via `addZarrLayer`, not just OCS, and also unlocks polygon-based zonal statistics for free.

giswqs:
Fixes #1555.

Implements option 1 from the issue: a passthrough on `GeoLibreAppAPI`, so a plugin that renders a cube with `addZarrLayer` can read its values.

## The gap

`addZarrLayer` returns only a layer id, and the `@carbonplan/zarr-layer` instance lives in module scope inside `maplibre-components.ts`, so `queryData` was unreachable on the native path. A plugin wanting click-to-value had to open the store a second time with its own zarrita point read, duplicating the reprojection and chunk indexing the renderer had already done.

## The change

`queryZarrLayer(layerId, geometry, selector?, options?)` is the read counterpart of `setZarrLayerSelector`: it looks the layer up in the control's layers map by id and delegates to the instance's `queryData`, resolving `null` for an id the Zarr renderer does not own (matching `setZarrLayerSelector` resolving `false`).

- A `Point` (WGS84, straight from a map click) gives click-to-value; a `Polygon`/`MultiPolygon` gives region statistics.
- `selector` scopes the read only, so an Identify readout for another `time` does not move the slice on screen.
- `options.signal` cancels a read the user has moved past; `includeSpatialCoordinates: false` drops the per-pixel coordinate arrays.
- Returned `coordinates` are in the store's **source** CRS, which the docs now call out.
- The geometry/selector/options/result types are aliased from the renderer's own types (`GeoLibreZarrQueryGeometry` and friends), so a change to that contract fails the host build rather than a hand-copied shape drifting silently, and a plugin does not need `@carbonplan/zarr-layer` as its own dependency.

Option 2 (wiring this into GeoLibre's own Identify tool, with host-owned presentation) is deliberately left out of scope: it is a UI decision on top of this primitive, and each plugin can now present the value in its own panel.

## Verification

Driven in the real web app (dev server) with a throwaway drop-in plugin standing in for the Open Climate Service plugin, against the public NOAA OI SST V2 monthly Zarr store (`sst`, 1-degree global grid), in **both light and dark themes**:

- Before the change: `typeof app.queryZarrLayer === "undefined"` while `setZarrLayerSelector` was already a function; the probe reported that click-to-value was impossible.
- After: a click at -127.891, 6.687 returned `{"sst":[27.42],"dimensions":["lat","lon"],"coordinates":{"lat":[6.5],"lon":[-127.5]}}` (source-CRS cell center on the 1-degree grid), and a 4-degree box around it returned 16 pixels, mean 27.33.
- A point outside the grid returned empty value arrays rather than an error; a `{ time: 400 }` override read a different slice without changing the rendered one; `includeSpatialCoordinates: false` dropped the coordinate arrays; an unknown layer id resolved `null`; an aborted query rejected.

Gates: `npm run build`, the full frontend suite (4337 pass, 3 new tests in `tests/zarr-layer-api.test.ts`), and pre-commit on the changed files.

### DIFF
--- a/apps/geolibre-desktop/src/hooks/usePlugins.ts
+++ b/apps/geolibre-desktop/src/hooks/usePlugins.ts
@@ -7,6 +7,7 @@ import {
   addRasterToMap,
   addZarrRasterLayer,
   buildSelectorTimeBinding,
+  queryZarrLayer,
   registerTemporalLayer,
   unregisterTemporalLayer,
   isTimeSliderIdle,
@@ -85,6 +86,9 @@ import type {
   GeoLibreTileLayerOptions,
   GeoLibreWmsLayerOptions,
   GeoLibreZarrLayerOptions,
+  GeoLibreZarrQueryGeometry,
+  GeoLibreZarrQueryOptions,
+  GeoLibreZarrQuerySelector,
 } from "@geolibre/plugins";
 import { invoke } from "@tauri-apps/api/core";
 import { open } from "@tauri-apps/plugin-dialog";
@@ -940,6 +944,15 @@ export function createAppAPI(mapControllerRef?: RefObject<MapController | null>)
       }),
     setZarrLayerSelector: (layerId: string, selector: Record<string, number | string>) =>
       setZarrLayerSelector(layerId, selector),
+    // Click-to-value and region statistics on a natively rendered cube: the
+    // renderer owns the grid, so it reprojects the WGS84 geometry and masks fill
+    // values itself instead of every plugin re-reading the store (#1555).
+    queryZarrLayer: (
+      layerId: string,
+      geometry: GeoLibreZarrQueryGeometry,
+      selector?: GeoLibreZarrQuerySelector,
+      options?: GeoLibreZarrQueryOptions,
+    ) => queryZarrLayer(layerId, geometry, selector, options),
     // A layer whose time is an internal dimension joins the Time Slider through
     // an adapter rather than a filter or a source swap. Registering only makes
     // it bindable; `bind` writes the binding and opens the dock, which is what a
--- a/docs/plugin-api.md
+++ b/docs/plugin-api.md
@@ -95,6 +95,13 @@ export interface GeoLibreAppAPI {
     layerId: string,
     selector: Record<string, number | string>,
   ) => Promise<boolean>;
+  // Click-to-value / region statistics on a Zarr layer (see "Zarr layers").
+  queryZarrLayer?: (
+    layerId: string,
+    geometry: GeoLibreZarrQueryGeometry,
+    selector?: GeoLibreZarrQuerySelector,
+    options?: GeoLibreZarrQueryOptions,
+  ) => Promise<GeoLibreZarrQueryResult | null>;
   // Register a layer the plugin added to the map itself, so it appears in the
   // Layers panel (see "Custom (WebGL) layers and paint ownership" below).
   registerExternalNativeLayer?: (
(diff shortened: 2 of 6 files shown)


giswqs:
@coderabbitai review

turban (commented):
Reviewed as the filer of #1555 — this implements option 1 faithfully. **LGTM.**

## Overview
`queryZarrLayer(layerId, geometry, selector?, options?)` delegates to the native `@carbonplan/zarr-layer` instance's `queryData`, unlocking click-to-value (`Point`) and zonal stats (`Polygon`/`MultiPolygon`) for cubes rendered via `addZarrLayer`.

## Correctness ✅
- **Faithful mirror of `setZarrLayerSelector`**: same `zarrControl?.getLayersMap().get(layerId)` lookup, same `!instance || typeof instance.queryData !== "function"` guard, `null`-for-unknown-id (vs the setter's `false`).
- **Read-only by construction**: unlike the setter it doesn't touch `useAppStore`, so a `selector` override reads another slice *without moving the display* — the key guarantee, and the test proves it (`selectorCalls.at(-1)` stays `undefined` after `{ time: 12 }`).
- **Geometry passed untouched** to the renderer (which owns the WGS84→source-CRS reprojection); the test guards against a future transform sneaking in on the way in.
- Signature matches `queryData`; abort rejection propagates as documented.

## Quality / conventions — strong
- **Type aliasing** (`GeoLibreZarrQuery*` ← the renderer's `QueryGeometry`/`Selector`/`QueryOptions`/`QueryResult`) is the right call: plugins annotate calls without depending on `@carbonplan/zarr-layer`, and a renderer-contract change fails the host build instead of drifting silently.
- Docs call out the easy footguns: source-CRS coordinates, empty-vs-`null`-vs-reject, and selector-scopes-read.

## Suggestions (minor, non-blocking)
- **Test gap**: the forward-compat branch — instance present but *without* `queryData` (an older renderer) → `null` — isn't exercised; only the unknown-id path is. A stub without `queryData` would cover it.
- **Result ergonomics**: `GeoLibreZarrQueryResult`'s `[variable: string]: unknown` index signature forces consumers to cast (`result.sst as number[]`). Faithful to the renderer contract, but a typed accessor or a prominent "you'll cast the variable array" note would smooth adoption.
- **Nit**: `plugin-api.md` hand-writes the query type shapes rather than pointing at the aliases — small drift risk vs the exported types.

## Risk
- No new surface (reads already-served data; geometry goes to the renderer, no injection path).
- A large polygon can pull many chunks, but that's inherent to zonal stats and cancelable via `options.signal` — fine for a primitive.

Thanks for turning this around so quickly — it's the thin `app.queryZarrLayer?.(...)` call we needed for click-to-value in the Open Climate Service plugin, no second zarrita read.

giswqs:
@turban Thank you for confirming.

Issue #1524, opened 2026-07-29
First-class map elements (pins, sticky notes, placed images) and an Elements panel


giswqs:
## Summary

Promote map annotations from decoration to first-class **map elements** — pins, sticky notes, and placed images with titles and descriptions — and add an **Elements** panel that lists everything on the map with click-to-zoom, rename, visibility, and delete.

## Why

The annotations plugin (`packages/plugins/src/plugins/maplibre-annotations.ts`) supports text, arrows, and highlights, and its own header says it plainly: "annotations are decoration". Management is limited to **Delete last** and **Clear all**, so once several annotations are on the map there is no way to find, rename, re-select, or individually remove one — only to undo in order or wipe the lot.

For explanatory maps — the ones people actually share — the annotation layer is the part that carries the meaning, and it is currently the least manageable part of the app.

## Proposal

**1. Richer element types**

- **Pin** — a placed marker with a title and description, using the existing built-in marker library and the data-driven icon sizing already in the style toolkit. Clicking a pin opens its description.
- **Sticky note** — a small text card anchored to a location, with a background color, that stays legible at any zoom.
- **Placed image** — drop an image onto the map at a location or corner-pinned to an extent. The image-overlay layer type and the Raster Georeferencer already provide both placement modes.
- Keep the existing text, arrow, and highlight tools as element types alongside these.

**2. An Elements panel**

A right-sidebar panel (registerable through the plugin UI host API) listing every element with its type glyph and title:

- Click to select and zoom.
- Rename inline.
- Per-element visibility toggle and delete.
- Reorder, so overlapping elements can be brought forward.

**3. Persistence**

Elements already save with the project; extend the stored shape to carry `title`, `description`, and `visible` per element, keeping the existing grouped-id mechanism (`annotationId`) that holds an arrow's parts together.

## Scope notes

- These are presentation elements, not data — they should stay distinct from GeoEditor-drawn features, which belong in real layers with attributes.
- Placed images need a size cap and should be embedded or referenced consistently with how the project format handles other binary payloads.

## Effort

Medium.

HarshShinde0:
@giswqs This issue has been resolved.  Can we close the issue?

PR #1567, opened 2026-07-30
feat: add Elements panel & promote annotations to map elements #1524

HarshShinde0:
map annotations from basic decorations to first-class map elements (Pins, Sticky Notes, and Placed Images) and adds an Elements Panel in the right sidebar to manage them.

image

## Summary by CodeRabbit

- **New Features**
  - Expanded annotations with **pin**, **sticky note**, and **placed image**, plus a right-side **Elements** panel to edit titles, reorder, toggle visibility, zoom to items, and delete them.
  - Added support for **point** and **extent/overlay** placed-image placement.
  - Introduced new localized prompts/labels for these tools and placement modes (with toolbar label fallback when translations are missing).
- **Bug Fixes**
  - Annotation layers now exclude features marked `visible: false`.
  - Generated images now fall back to a transparent placeholder when generation fails, and placeholders can be safely replaced.
- **Tests**
  - Added Elements Panel and element/overlay synchronization tests, including reorder behavior.

### DIFF
--- a/apps/geolibre-desktop/src/components/layout/TopToolbar.tsx
+++ b/apps/geolibre-desktop/src/components/layout/TopToolbar.tsx
@@ -200,12 +200,16 @@ export function TopToolbar({
     setAnnotationLabels({
       toolbar: t("annotations.toolbar"),
       layerName: t("annotations.layerName"),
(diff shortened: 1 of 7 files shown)


giswqs (commented):
This is great! Thank you for your contribution.

PR #1654, opened 2026-08-02
fix(backend): validate aspatial geopackage layer writes & add test co…

RohithPariki:
### What
`_write_geopackage` accepted any layer name that existed in the GeoPackage, including aspatial tables (those with `geometry_type = None`). Passing an aspatial table name caused an OGR crash at write time instead of a clean HTTP error

### Why
The existing guard only checked `if layer not in names` (404 for missing), but didnt check whether the matched layer was actually a spatial table.

### Fix
Add a second check `if layer not in spatial` after  404 guard, returning 400 with clear message: `"Layer 'X' is an aspatial table and cannot be written"`.

### Tests
- `test_write_geopackage_rejects_aspatial_layer` - verifies the new 400 path
- `test_fix_geometries_leaves_unfixable_unchanged` -  covers the previously untested `unfixable` branch in `_fix_geometries` (geometry that `make_valid` cant repair is left unchanged and counted)

All 119 existing + new vector tests pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * GeoPackage write-back now rejects explicitly selected non-spatial tables with a clear HTTP 400 error.
  * GeoPackages containing no spatial layers now return a clear HTTP 400 error.
  * Missing layers continue to return HTTP 404 errors.
  * Geometry repair now preserves unrepairable degenerate polygons and reports them accurately.

### DIFF
--- a/backend/geolibre_server/geolibre_server/app/vector.py
+++ b/backend/geolibre_server/geolibre_server/app/vector.py
@@ -229,6 +229,11 @@ def _write_geopackage(target: Path, geojson: dict, layer: Optional[str]) -> tupl
                 status_code=404,
                 detail=f"Layer '{layer}' not found in {target.name}",
             )
+        if layer not in spatial:
+            raise HTTPException(
+                status_code=400,
+                detail=f"Layer '{layer}' is an aspatial table and cannot be written",
+            )
         target_layer = layer
     elif spatial:
         # Match the reader's "first feature layer" default (gpkg-reader.ts
--- a/backend/geolibre_server/geolibre_server/app/vector.py
+++ b/backend/geolibre_server/geolibre_server/app/vector.py
@@ -214,8 +214,8 @@ def _write_geopackage(target: Path, geojson: dict, layer: Optional[str]) -> tupl
         A ``(feature_count, layer_name)`` tuple.
 
     Raises:
-        HTTPException: The named layer is absent, or the file has several feature
-            layers and none was specified.
+        HTTPException: The named layer is absent, the named layer is an aspatial
+            attribute table, or the file holds no feature layer at all.
     """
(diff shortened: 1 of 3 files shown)


giswqs:
Pushed b12ffbb2 addressing the review feedback — thanks @RohithPariki for catching the aspatial write path in the first place.

**The major one** ([`vector.py:232-236`](https://github.com/opengeos/GeoLibre/pull/1654#discussion_r1)): the new explicit-layer guard was right, but the auto-select path had the same hole. With `layer=None`, a GeoPackage whose only table is aspatial fell through to `elif len(names) == 1: target_layer = names[0]` and handed OGR exactly the table the guard above rejects. That fallback is redundant for every other case — a lone spatial layer is already covered by `elif spatial: target_layer = spatial[0]` — so it is gone, and the existing `No feature layer to write` 400 now covers it.

Verified rather than assumed: the new `test_write_geopackage_rejects_aspatial_only_file` fails on the pre-fix code with `pyogrio.errors.DataLayerError: Layer 'aspatial_data' could not be opened`, and passes after.

**The rest:**
- `test_vector.py` — assert the complete `exc.value.detail` instead of the `"aspatial table"` substring; `mock_list_layers(_path: str) -> pd.DataFrame` (ANN202/ARG001).
- `test_vector_ops.py` — `test_fix_geometries_leaves_unfixable_unchanged` is now parametrized over `raises` / `returns_empty`, so both branches of `_repair` are exercised (`make_valid` raising, and returning an empty geometry); mock annotated `(_geom: object) -> NoReturn` with a bare `raise ValueError`.
- Extended the `Raises:` docstring on `_write_geopackage` to mention the aspatial-table 400 (@github-actions' nit).

Backend suite: 372 passed, 23 skipped. `pre-commit run --files` on the three changed paths is clean.

PR #1703, opened 2026-08-04
fix(map): correct basemap symbol occlusion under draped rasters in 3D…

### LINKED ISSUE #781: UI Bug Report: Basemap Layer Incorrectly Renders in Foreground of Overlying Raster Layers With Terrain Plugin also active
Environment Details:

Environment: Mac Osx 15.7
Browser: Safari 26.2 (20623.1.14.18.4)
GeoLibre version: 1.7.0

---

### Description

The layer stack hierarchy is not being respected when rendering vector basemap elements over raster layers. Specifically, the layer named **Background** (designated as a BASEMAP) incorrectly overlays its features on top of user-added RASTER layers positioned higher in the list, causing visual clipping and data obstruction.

----

### Note
This unexpected behavior **only surfaces when the Terrain plugin is active** in the project workspace, suggesting it might be an intentional design choice or a limitation tied to 3D rendering pipelines rather than a standard 2D layout bug.

This issue is clearly demonstrated across two distinct user interface states shown below:

* **Example 1 (Satellite Imagery Overlap):** In the first user interface shown below, the layer **Esri Wayback 2014-02-20** is at the top of the stack and marked as visible. The **Background** layer sits at the bottom of the stack. However, as indicated by the red arrow on the map, vector labels and building footprints from the basemap, such as **The West Wing**, **West Colonnade**, and **White House**, are rendered directly on top of the aerial photography.

* **Example 2 (Topographic Map Overlap):** In the second user interface shown below, the top layer is hidden, and the middle layer, **USGS USGS Topo**, is activated. The same behavior occurs: the underlying **Background** basemap details cut through and display on top of the topographic map lines, green park boundaries, and elevation contours.

---

### Proposed User Experience

Users expect a predictable, bottom-up rendering order where elements higher in the **Layers** list completely obscure or stack on top of the elements below them, depending on opacity settings.

To resolve this issue, the interface should behave as follows:

* When a raster layer like **Esri Wayback 2014-02-20** or **USGS USGS Topo** is positioned above the **Background** layer in the stack and set to 1.00 opacity, it must completely cover all vector structures, labels, and polygons belonging to that basemap.
* Vector labels and features from the basemap should only become visible if the user lowers the opacity slider of the overlying raster layer, or manually drags the basemap layer to a higher position in the list.

----

Image

Image

RohithPariki:
## Summary
This PR fixes a visual bug where vector basemap labels and symbols were incorrectly rendering on top of overlying opaque raster layers when the 3D Terrain plugin was active.

## Problem
In Issue #781, users observed that the layer stack hierarchy was broken when using 3D terrain. A designated "Background" basemap layer would incorrectly draw its labels, buildings, and vector footprints directly over top-level, fully opaque user rasters (e.g., Esri Wayback, USGS Topo).

## Root Cause
This was caused by a known rendering limitation in our underlying engine, `maplibre-gl-js`. In 3D terrain mode, MapLibre uses a two-pass rendering system: draped layers (rasters, fills) are drawn to a texture and draped on the terrain mesh, while non-draped symbols and extrusions are drawn "live" in a separate pass afterwards to avoid clipping. This decoupled rendering meant that all basemap symbols floated above the previously drawn rasters, effectively ignoring the 2D layer list hierarchy.

## Solution
Instead of attempting a fragile workaround within GeoLibre's UI layer-sync logic, this PR applies a surgical patch directly to our `maplibre-gl` dependency using `patch-package`. The patch backports an open upstream fix (MapLibre PR #7852) which modifies `RenderToTexture.ts` to identify the highest opaque draped raster layer and explicitly skips rendering for symbol layers placed below it.

*Testing Note:* Verified locally on Windows. The frontend UI test suite passes successfully. The 8 subtest failures observed locally were confirmed to be WSL/bash path resolution issues in `metainfo-generated.test.ts` (Linux AppImage packaging) running on a Windows host, and are unrelated to this UI fix.

## References
Closes #781

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved map rendering by preventing symbols and 3D fill features from appearing beneath fully opaque raster layers.
  * Added coverage for raster opacity scenarios to improve visual consistency.

* **Chores**
  * Installation now automatically applies the required map rendering compatibility patch.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,6 +7,7 @@
     "": {
       "name": "geolibre",
       "version": "2.4.1",
+      "hasInstallScript": true,
       "workspaces": [
         "apps/*",
         "packages/*",
@@ -19,6 +20,7 @@
         "eslint": "^10.8.0",
         "eslint-plugin-react-hooks": "^7.1.1",
         "linkedom": "^0.18.13",
+        "patch-package": "^8.0.0",
         "tsx": "^4.23.1",
         "typescript-eslint": "^8.65.0"
       },
@@ -4211,9 +4213,6 @@
         "arm"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
@@ -4231,9 +4230,6 @@
         "arm64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
@@ -4251,9 +4247,6 @@
         "ppc64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
@@ -4271,9 +4264,6 @@
         "riscv64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
@@ -4291,9 +4281,6 @@
         "s390x"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
@@ -4311,9 +4298,6 @@
         "x64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "LGPL-3.0-or-later",
       "optional": true,
       "os": [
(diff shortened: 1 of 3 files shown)


giswqs:
@RohithPariki Thank you for looking into it. The issue is being fixed in the upstream. I would prefer not to apply a surgical patch in GeoLibre. 

https://github.com/maplibre/maplibre-gl-js/issues/7828
https://github.com/maplibre/maplibre-gl-js/pull/7852

RohithPariki:
Makes complete sense! I'll close this PR and we can wait for the upstream MapLibre release to bump the dependency.
 Thanks for the quick review

PR #1781, opened 2026-08-08
feat(i18n): use Vazirmatn font for Persian locale

RMA1313:
## Summary

Use Vazirmatn as the UI font for the Persian (`fa`) locale.

IBM Plex Sans does not provide Persian glyph coverage, so Persian text currently falls back to the system font. This change adds Vazirmatn as a self-hosted font and applies it only when the active locale is Persian.

## Changes

- Add `@fontsource-variable/vazirmatn`
- Load Vazirmatn using the existing self-hosted font setup
- Apply Vazirmatn only to the Persian (`fa`) locale
- Keep the existing IBM Plex Sans font stack unchanged for other locales

## Testing

- `npm run build -w geolibre-desktop` ✅

## Summary by CodeRabbit

* **Style**
  * Added improved Persian and Arabic typography using the self-hosted Vazirmatn variable font.
  * Updated the font stack to prioritize Vazirmatn while retaining existing and system fallback fonts.
  * Clarified how application font stylesheets are loaded.

### DIFF
--- a/apps/geolibre-desktop/package.json
+++ b/apps/geolibre-desktop/package.json
@@ -32,6 +32,7 @@
     "@electric-sql/pglite": "^0.5.4",
     "@electric-sql/pglite-postgis": "^0.2.4",
     "@fontsource-variable/ibm-plex-sans": "^5.3.0",
+    "@fontsource-variable/vazirmatn": "^5.3.0",
     "@fontsource/ibm-plex-mono": "^5.3.0",
     "@geolibre/core": "*",
     "@geolibre/embed": "*",
--- a/apps/geolibre-desktop/src/index.css
+++ b/apps/geolibre-desktop/src/index.css
@@ -27,6 +27,13 @@
     "IBM Plex Mono", ui-monospace, SFMono-Regular, Menlo, Consolas, "Liberation Mono", monospace;
 }
 
+html:lang(fa) {
+  --font-sans:
+    "Vazirmatn Variable", ui-sans-serif, system-ui, -apple-system, "Segoe UI", Roboto,
+    "Noto Sans Arabic", "Noto Sans", "Helvetica Neue", Arial, sans-serif,
+    "Apple Color Emoji", "Segoe UI Emoji";
+}
+
 /* Map-floating chrome, a.k.a. "instrument glass".
 
    The distinction this encodes: chrome docked BESIDE the map (top toolbar,
--- a/apps/geolibre-desktop/src/index.css
+++ b/apps/geolibre-desktop/src/index.css
@@ -30,8 +30,8 @@
 html:lang(fa) {
   --font-sans:
     "Vazirmatn Variable", ui-sans-serif, system-ui, -apple-system, "Segoe UI", Roboto,
-    "Noto Sans Arabic", "Noto Sans", "Helvetica Neue", Arial, sans-serif,
-    "Apple Color Emoji", "Segoe UI Emoji";
+    "Noto Sans Arabic", "Noto Sans", "Helvetica Neue", Arial, sans-serif, "Apple Color Emoji",
+    "Segoe UI Emoji";
 }
 
 /* Map-floating chrome, a.k.a. "instrument glass".
--- a/apps/geolibre-desktop/src/index.css
+++ b/apps/geolibre-desktop/src/index.css
@@ -7,10 +7,10 @@
 
 /* App typeface: IBM Plex Sans for UI, IBM Plex Mono for the numeric/coordinate
    readouts (StatusBar, code, feature IDs). One matched superfamily, so the mono
-   readouts and the surrounding UI share a voice. The @font-face rules are
-   loaded from main.tsx — see the note there for why they cannot be @imported
-   here. Self-hosted rather than CDN-loaded: the desktop build must render
-   offline, and the Tauri CSP is `default-src 'self'`.
+   readouts and the surrounding UI share a voice. Fontsource's own @font-face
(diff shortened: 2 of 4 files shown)


giswqs:
Nice fix — Persian genuinely had no glyph coverage before this. One suggestion on the bundle cost, with numbers from building `main` and this branch back to back.

### The cost today

`@fontsource-variable/vazirmatn/wght.css` declares **all three** of the font's subsets, so Vite emits all three:

| asset | bytes |
|---|---|
| `vazirmatn-arabic-wght-normal.woff2` | 46,308 |
| `vazirmatn-latin-wght-normal.woff2` | 34,524 |
| `vazirmatn-latin-ext-wght-normal.woff2` | 21,860 |
| **total** | **102,692** |

Whole-`dist` delta is **+105,462 B** (102,692 of woff2, +1,931 B of `@font-face` CSS in `index-*.css`, and the rest precache-manifest growth). No JS change.

The part worth flagging is the service worker: the three files land in the Workbox precache manifest, which goes from **458 entries / 15,707.43 KiB** to **461 / 15,809.35 KiB** — **+101.92 KiB**. Precaching fetches every entry on SW install, so this isn't lazy per-`unicode-range` font loading; once the SW registers, *every* user pays the full ~100 KiB whether or not they ever select Persian.

### Suggestion: load only the Arabic subset

Persian only needs the `arabic` subset. The two Latin subsets are 56,384 B — 55% of the added weight — and the app already ships IBM Plex Sans for Latin.

There's no subset-scoped entry point to import (the package publishes only `index.css` and `wght.css`, both listing all three subsets), but `./files/*.woff2` is exported, so a hand-written `@font-face` in `index.css` works. Drop the `main.tsx` import and add:

```css
@font-face {
  font-family: "Vazirmatn Variable";
  font-style: normal;
  font-display: swap;
  font-weight: 100 900;
  src: url("@fontsource-variable/vazirmatn/files/vazirmatn-arabic-wght-normal.woff2")
    format("woff2-variations");
  unicode-range: U+0600-06FF,U+0750-077F,U+0870-088E,U+0890-0891,U+0897-08E1,U+08E3-08FF,U+200C-200E,U+2010-2011,U+204F,U+2E41,U+FB50-FDFF,U+FE70-FE74,U+FE76-FEFC,U+102E0-102FB,U+10E60-10E7E,U+10EC2-10EC4,U+10EFC-10EFF,U+1EE00-1EE03,U+1EE05-1EE1F,U+1EE21-1EE22,U+1EE24,U+1EE27,U+1EE29-1EE32,U+1EE34-1EE37,U+1EE39,U+1EE3B,U+1EE42,U+1EE47,U+1EE49,U+1EE4B,U+1EE4D-1EE4F,U+1EE51-1EE52,U+1EE54,U+1EE57,U+1EE59,U+1EE5B,U+1EE5D,U+1EE5F,U+1EE61-1EE62,U+1EE64,U+1EE67-1EE6A,U+1EE6C-1EE72,U+1EE74-1EE77,U+1EE79-1EE7C,U+1EE7E,U+1EE80-1EE89,U+1EE8B-1EE9B,U+1EEA1-1EEA3,U+1EEA5-1EEA9,U+1EEAB-1EEBB,U+1EEF0-1EEF1;
}
```

(`unicode-range` copied verbatim from the package's `wght.css` — worth keeping it exact.) The npm dependency stays; only the CSS entry point changes.

I built this variant to check it against the Tailwind v4 hazard documented above the font imports in `main.tsx` — that footgun is specific to `@import`ing fontsource's own CSS with its relative `url(./files/*.woff2)`. A hand-written `@font-face` in `index.css` using a bare package specifier is rewritten by Vite normally, and exactly one file is emitted:

```
dist/assets/vazirmatn-arabic-wght-normal-Cafbb7Zc.woff2   46,308 B
```

| | dist delta vs `main` | precache delta |
|---|---|---|
| this PR | +105,462 B | +101.92 KiB (3 entries) |
| arabic-only | **+47,477 B** | **+45.44 KiB (1 entry)** |

Saves **57,985 B** of shipped assets and **56.48 KiB** of precache.

### One thing to pair with it

The `html:lang(fa)` stack currently goes `"Vazirmatn Variable", ui-sans-serif, system-ui, …` — it doesn't include IBM Plex Sans. With all three subsets loaded, Latin text in the Persian UI renders in Vazirmatn Latin; with the Arabic subset only, it would fall through to `ui-sans-serif`/`system-ui` and no longer match the rest of the app. Adding `"IBM Plex Sans"` right after Vazirmatn fixes that at zero extra bytes, since Plex is already loaded:

```css
--font-sans:
  "Vazirmatn Variable", "IBM Plex Sans", ui-sans-serif, system-ui, …;
```

That arguably makes Persian *more* consistent with every other locale than the current version does, on top of being smaller.

giswqs:
Pushed this to the branch as [`67bbb91`](https://github.com/opengeos/GeoLibre/pull/1781/commits/67bbb91e) rather than leaving it as a review note — hope that's alright, and thanks for the PR.

Verified in a real browser against the production build (`?locale=fa`):

- `<html lang="fa" dir="rtl">`, and Persian text (`پروژه`) computes to `"Vazirmatn Variable", "IBM Plex Sans Variable", …`
- `document.fonts.check('16px "Vazirmatn Variable"')` → `true`
- exactly three font requests, all 200: `vazirmatn-arabic-wght-normal`, `ibm-plex-sans-latin-wght-normal`, `ibm-plex-mono-latin-400-normal` — the Latin fallback resolves to Plex as intended, so the status bar and coordinate readouts keep the same typeface as every other locale

`npm run build`, the frontend suite (5,499 passing), and `pre-commit` are all green.

PR #1784, opened 2026-08-08
feat: expose read-only layer queries to external plugins

Jihkk:
## Summary

- expose optional read-only layer and feature query methods to external plugins
- expose the selected layer, all selected features, sketch features, and selection-change subscriptions
- document compatibility and cleanup behavior for plugin authors
- add focused tests for selection, layer lookup, sketch filtering, missing IDs, and unsubscribe behavior

## Why

External plugins can add layers, but they currently cannot inspect GeoLibre's current layer list or feature selection through the public plugin API. Plugins that need to run an analysis for a user-selected feature therefore have no supported read-only path to obtain that context.

The new methods remain optional so existing plugins can continue to run against older GeoLibre hosts. The query surface does not mutate the application store, and subscription callers receive an unsubscribe function for plugin cleanup.

## API

- `listLayers()`
- `getLayerFeatures(layerId)`
- `getSelectedFeatures()`
- `getSelectedLayerId()`
- `getDrawnFeatures()`
- `onSelectionChange(callback)`

## Validation

- `node --import tsx --test tests/plugin-query-api.test.ts` — 7 passed
- `npm run lint` — 0 errors (40 existing warnings)
- `npm run build` — passed

The API was also exercised from an external hydrology plugin in the Tauri desktop during a synthetic end-to-end workflow.

## Summary by CodeRabbit

* **New Features**
  * Added read-only plugin APIs for listing layers and retrieving layer, selected, and drawn features.
  * Plugins can now identify the selected layer and subscribe to selection changes.
  * Added layer visibility, opacity, type, and name details to layer information.
  * Documented selection behavior, empty results, error handling, and subscription cleanup.

* **Tests**
  * Added coverage for layer and feature queries, selection updates, drawn features, errors, and immutable results.

PR #1890, opened 2026-08-14
fix(ci): re-measure a line-only coverage shortfall before failing

### LINKED ISSUE #1889: Frontend line-coverage floor is nondeterministic on CI and can fail spuriously
## Summary

`npm run test:frontend:coverage` gates on `--test-coverage-lines=78`. That **line** number is not reproducible on CI: two runs over byte-identical sources produced **81.82%** and **76.47%**, and the second is below the floor. Branch and function coverage were identical in both.

This means the line floor can redden a PR that changed nothing relevant. It already did: PR #1887 (a release version bump plus docs, zero source files touched) failed with

```
Error: 76.47% line coverage does not meet threshold of 78%.
```

while `main`'s own run of the same sources reported 81.82% and did not.

## Evidence

Two CI runs, same TypeScript sources (the PR differed only in `package.json` versions, docs, and lockfiles):

| run | files counted | lines | branches | functions |
|---|---|---|---|---|
| `main` push ([run 31764040051](https://github.com/opengeos/GeoLibre/actions/runs/31764040051)) | 444 | **81.82%** | 84.30% | 60.36% |
| PR #1887 ([run 31764356629](https://github.com/opengeos/GeoLibre/actions/runs/31764356629)) | 444 | **76.47%** | 84.30% | 60.36% |

Both executed identically: 5935 tests, 1297 suites, 5934 pass, 0 fail, 1 skipped, 0 todo.

Comparing the two per-file tables by full path:

- **114 of 444 files** report different line coverage
- **0 files** report different function coverage
- **0 files** report different branch coverage
- the counted file set is identical (no file entered or left the report)

The swings are large, not rounding noise:

| file | run A | run B |
|---|---|---|
| `packages/core/src/storymap-sample.ts` | 99.21% | 34.65% |
| `packages/map/src/qml-import.ts` | 97.53% | 48.48% |
| `packages/map/src/qml-export.ts` | 94.38% | 44.52% |
| `packages/core/src/visibility.ts` | 93.42% | 46.05% |
| `packages/map/src/sld-export.ts` | 97.70% | 54.75% |
| `packages/core/src/layer-ref-scrub.ts` | 97.96% | 57.14% |

Affected files cluster in `packages/plugins` (46), `packages/core` (37), `packages/map` (16), `apps/geolibre-desktop` (9). Some are referenced by many test files (`visibility.ts` by 33); others (`storymap-sample.ts`, `layer-ref-scrub.ts`) by none directly, reaching the report only transitively through a package index.

## Not reproducible locally

Two consecutive local runs on one tree differed in **1** file (`packages/core/src/store.ts`, 75.35% vs 76.36%), totals 82.84% vs 82.86%. So this looks specific to the CI runner, plausibly `node --test`'s worker-process count and the merging of per-process V8 coverage under a 4-core, memory-constrained runner. That functions and branches are perfectly stable while lines are not suggests the line ranges, rather than the coverage data as a whole, are what get lost or partially merged.

## Why it matters

Per `CLAUDE.md` the floors are a ratchet meant to catch real regressions. A gate that swings 5+ points run to run on identical code cannot do that: it produces failures nobody can act on, and it trains people to re-run CI until it passes, which is exactly how a genuine regression slips through.

## Options

1. Lower the line floor to sit under the observed trough (the swing bottomed at 76.47%, so ~74) and keep functions and branches where they are, since those are stable.
2. Drop the line gate and ratchet on functions and branches only, which are reproducible run to run.
3. Pin `node --test` concurrency in CI (`--test-concurrency=1` or a small fixed number) and re-measure. Slower, but if it makes lines deterministic the floor can stay meaningful.
4. Report upstream to Node if a minimal reproduction can be extracted.

I lean toward 3 for a measurement, then 1 or 2 depending on what it shows.

## Context

Found while diagnosing the separate, real coverage regression fixed in #1888. That one was deterministic (function coverage, 72.90% to 60.36%) and is unrelated to this flake.

giswqs:
Fixes #1889.

## The problem

Line coverage is nondeterministic on CI. Two runs over byte-identical sources:

| run | files | lines | branches | functions |
|---|---|---|---|---|
| `main` push | 444 | **81.82%** | 84.30% | 60.36% |
| PR #1887 | 444 | **76.47%** | 84.30% | 60.36% |

Same 5935 tests, 1297 suites, 5934 pass, 0 fail, 1 skipped in both. Comparing the per-file tables by full path: **114 of 444 files differ on lines, zero differ on branches or functions**, and no file entered or left the report. Swings are large, not rounding: `storymap-sample.ts` 99.21% to 34.65%, `qml-import.ts` 97.53% to 48.48%.

The practical result is that PR #1887, a release bump touching only version strings, docs, and lockfiles, failed a floor that the identical sources had cleared minutes earlier.

## What I ruled out

Not reproducible locally, across 11 runs:

- Node 26.4.0 (local default) and Node **22.23.2**, the exact version all four CI runs used: 82.81% three times running
- 24 cores, and pinned to 4 with `taskset` to match the runner: 82.84% twice
- default `--test-concurrency` vs `--test-concurrency=1`: identical numbers (82.80 / 84.46 / 72.94), at 107s serial vs 47s parallel

So pinning concurrency costs 2.3x wall time and buys nothing measurable, and I could not manufacture the collapse to verify a deterministic fix against it. Branch and function coverage have never varied in any observation, CI or local.

## The fix

Mitigate narrowly rather than lower the floor. `test:frontend:coverage` now runs through `scripts/coverage-check.mjs`, following the `scripts/audit-check.mjs` precedent of wrapping a tool that cannot express the policy we need.

Node still enforces all three floors. The wrapper only decides whether a failure deserves a second opinion:

| failure | behavior |
|---|---|
| line floor only, all tests passed | re-measure once; fail only if short again |
| branch or function floor | fail immediately, no retry |
| any test failure | fail immediately, no retry |
| non-zero exit with no threshold reported (crash, OOM) | fail immediately, no retry |

A real line regression reproduces and so fails on the second run. Nothing is retried away, and the deterministic metrics keep failing fast.

## Why the retry logic is tested

A retry that quietly swallowed a genuine regression would be worse than the flake it works around, so `classify()` is exported and pinned by `tests/coverage-check.test.ts` (8 cases, including the exact output shape that reddened #1887 and the exact one from the #1784 function regression). It also covers the `ℹ` vs `#` reporter prefixes, since Node switches on whether stdout is a TTY.

Verified end to end with impossible floors:

- `LINES=99`: suite runs **twice** (34s), logs both readings, exits 1
- `FUNCTIONS=99`: suite runs **once** (18s), exits 1, no retry
- unmodified: exits 0, one run, 82.79% / 84.45% / 72.96%

The one path not exercised end to end is "retry recovers, exit 0", because I cannot reproduce the flake on demand. Its two components are covered separately: the retry executes (shown above) and `classify().ok` on a clean run is unit-tested.

## Also in CLAUDE.md

Documented the wrapper, and separately the denominator trap behind the #1784 regression: writing the *first* test for a large untested module reads as a coverage drop, because the module and everything it imports enter the denominator at once. The fix is to test against a leaf module (as #1888 did), not to lower a floor. That one will recur, so it is worth writing down.

## If the retry starts firing regularly

It logs both numbers precisely so #1889 can accumulate evidence. Frequent retries are the signal to fix the measurement rather than widen the mitigation, and I have left #1889 open for that.

## Summary by CodeRabbit

* **Testing**
  * Improved frontend coverage validation with line, branch, and function thresholds.
  * Added a limited retry for line-coverage shortfalls when all tests pass.
  * Coverage failures involving branches, functions, test failures, or other errors now fail immediately with clearer reporting.
  * Added tests for validation outcomes, coverage reporting, and retry behavior.

* **Documentation**
  * Expanded guidance for interpreting coverage changes and addressing measurement issues.
  * Documented how to update tests when coverage retry behavior changes.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -42,7 +42,31 @@ a few points under the current numbers as a **ratchet** — regressions fail CI,
 and when coverage rises comfortably above a floor, raise the floor to lock in the
 gain. The frontend
 report only counts files a test actually imports, so a module with no test does
-not appear at all rather than as 0%. The backend coverage run (and `npm run ci`,
+not appear at all rather than as 0%.
+
+That last point is the one that bites: writing the *first* test for a large
+untested module reads as a coverage **regression**, because the module and
+everything it imports enter the denominator at once. GeoLibre#1784 added a test
+that imported `usePlugins.ts` and so pulled in the whole built-in plugin
+registry, 39 files, dropping function coverage 72.90% → 60.36% and reddening
+`main`. The fix is to test against a leaf module rather than to lower the floor
+(GeoLibre#1888 extracted `lib/plugin-layer-queries.ts`; `geo-editor-geometry.ts`
+in `@geolibre/plugins` is the same pattern). Check what a new test *transitively*
+imports before assuming a coverage drop means the code got worse.
+
+`test:frontend:coverage` runs through `scripts/coverage-check.mjs` rather than
+calling `node --test` directly. Node still enforces all three floors; the wrapper
+only re-measures once when **line** coverage alone comes up short with every test
+passing. Line coverage is nondeterministic on CI (GeoLibre#1889: two runs over
+byte-identical sources reported 81.82% and 76.47%, 114 of 444 files differing on
+lines and *none* on branches or functions), and it is not reproducible locally on
+either Node 22 or 26. Branch and function shortfalls, and any test failure, fail
+on the spot with no retry, so a real regression still fails fast. `classify()` is
+exported and covered by `tests/coverage-check.test.ts` — change the retry policy
+there, not by loosening a floor. If the retry starts firing regularly, fix the
+measurement instead of widening the mitigation.
+
+The backend coverage run (and `npm run ci`,
 which calls the `:coverage` variants) needs `pytest-cov` from the backend `dev`
 extra. Install the **`test`** extra to run the *full* backend suite — without
 the optional engines (geopandas/rasterio/sedona/httpx) the vector/raster/SQL/ML
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "lint": "eslint apps packages workers tests",
     "test": "npm run test:frontend",
     "test:frontend": "node --import tsx --test tests/*.test.ts",
-    "test:frontend:coverage": "node --import tsx --test --experimental-test-coverage --test-coverage-lines=78 --test-coverage-branches=78 --test-coverage-functions=63 --test-coverage-exclude=\"tests/**\" --test-coverage-exclude=\"e2e/**\" --test-coverage-exclude=\"**/*.config.*\" tests/*.test.ts",
+    "test:frontend:coverage": "node scripts/coverage-check.mjs",
     "test:backend": "python -m pytest backend/geolibre_server/tests",
     "test:backend:coverage": "python -m pytest backend/geolibre_server/tests --cov=geolibre_server --cov-report=term-missing --cov-fail-under=55",
     "test:worker": "npm run typecheck -w geolibre-viewer-worker && npm run typecheck -w geolibre-collab-worker && npm run typecheck -w geolibre-collab-node && npm test -w geolibre-collab-node && npm run typecheck -w geolibre-tiles-worker && npm run typecheck -w geolibre-ai-proxy-worker",
--- a/scripts/coverage-check.mjs
+++ b/scripts/coverage-check.mjs
@@ -1,0 +1,158 @@
+// CI frontend coverage gate: the `node --test` coverage floors, with one retry
+// reserved for the line metric.
+// Run with: node scripts/coverage-check.mjs  (or `npm run test:frontend:coverage`)
+//
+// Node enforces all three floors itself; this wrapper only decides whether a
+// failure is worth a second opinion. Line coverage is the one metric observed
+// to be nondeterministic on CI: two runs over byte-identical sources reported
+// 81.82% and 76.47%, with 114 of 444 files differing on lines while *zero*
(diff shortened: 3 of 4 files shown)


giswqs:
Pushed a follow-up commit fixing a bug in the wrapper that the first CI run exposed.

**The first version lost most of the CI log.** It captured the run with `spawnSync` and wrote the output afterwards, then called `process.exit`. `process.stdout` is asynchronous when it is a **pipe**, which is what a CI runner gives it, so the exit discarded everything still queued. The run for this branch went green while the log was cut off mid-line:

```
    ok 3 - honors explicit and default profiles over th
> geolibre@2.6.0 test:worker
```

About 42,000 lines of test output and the entire coverage summary were gone, and the step still exited 0. It passed locally first because redirecting to a file makes `process.stdout` synchronous, so the failure mode does not exist there.

**Fix:** stream the child's stdout and stderr through as they arrive, and return an exit code the caller assigns to `process.exitCode` rather than calling `process.exit` anywhere, so Node exits only after the output drains. Classification is unchanged, since it reads the accumulated string either way.

**Verified with stdout on a pipe** rather than the file redirect that hid it:

| | lines in log | ends with | summary |
|---|---|---|---|
| before | 2,044 | truncated mid-line | missing |
| after | 45,384 | `# end of coverage report` | intact |

Retry paths re-checked against the rewritten async code with impossible floors: `LINES=99` runs the suite twice and exits 1, `FUNCTIONS=99` runs it once and exits 1, unmodified exits 0 at 82.78% / 84.45% / 72.94%.

Also added a ninth test asserting the script never calls `process.exit`, confirmed to fail against the reintroduced bug. A truncated log looks green, so nothing downstream would catch this on its own.

PR #1940, opened 2026-08-15
fix(layout): persist Browser and Comments panel toggles

### LINKED ISSUE #1935: Desktop: Layout settings do not persist after restart when reopening last project
## Description

Layout preferences under **Settings → Layout** do not appear to persist across application restarts in GeoLibre Desktop.

For example, if I disable **Show Comments panel**, the Comments panel is hidden for the current session. However, after quitting and reopening GeoLibre Desktop, the Comments panel appears again, as if the saved layout preference has been reset to the default configuration.

My startup behavior is configured to **Reopen the last project**, so I would expect the saved layout preferences for that project/session to be restored as well.

## Steps to reproduce

1. Open GeoLibre Desktop.
2. Set startup behavior to **Reopen the last project**.
3. Go to **Settings → Layout**.
4. Disable **Show Comments panel**.
5. Confirm that the Comments panel disappears.
6. Quit GeoLibre Desktop.
7. Reopen GeoLibre Desktop and let it reopen the last project.

## Actual behavior

The Comments panel appears again after restart. The Layout setting seems to have reverted to its default state.

## Expected behavior

Layout preferences such as **Show Comments panel**, **Show Layers panel**, **Show Browser panel**, toolbar labels, etc. should persist across application restarts, especially when the startup behavior is set to **Reopen the last project**.

## Environment

- GeoLibre Desktop
- macOS
- Startup setting: **Reopen the last project**

## Additional context

This may indicate that the saved Layout settings are being overwritten or reinitialized from the default configuration during application/project startup.

giswqs:
Fixes #1935

## Problem

The **Settings → Layout** toggles for the **Browser** and **Comments** panels acted only on the live right-panel registry. Nothing was written to `desktopSettings.layout`, and both registration hooks unconditionally called `openRightPanel(...)` on mount, so every launch reopened a panel the user had turned off.

The startup mode was a red herring: "Reopen the last project" restores layers and the camera, never these two panels. The other Layout settings (Layers panel, Style panel, toolbar labels, project info) were already persisted and did survive a restart.

## Fix

- `DesktopLayoutSettings` gains `browserPanelVisible` and `commentsPanelVisible`, both defaulting to `true` and normalized with the same strict-boolean rule as the existing keys, so settings saved before these keys existed keep today's behavior.
- `useRegisterBrowserPanel` / `useRegisterCommentsPanel` seed the panel from that setting on mount instead of always opening it. They read the store rather than subscribing, so closing a panel from its own header is still a session-level action and is not written back as a preference.
- The Settings toggles write the setting alongside the registry call. They apply live rather than on Save, so they patch the dialog draft too: the draft is snapshotted when the dialog opens and Save writes `layout` wholesale, which would otherwise revert the toggle the user just made in that same dialog.
- Layout **Reset** now moves the two panels as well, since those rows render the live registry state rather than the draft.

## Verification

Driven in the real app with Playwright, with `us_cities.geojson` loaded, in **both dark and light themes**:

- Fresh settings: both rails present (defaults unchanged).
- Toggle Comments off from the Settings → Layout submenu → `commentsPanelVisible: false` in storage; after reload the Comments rail is gone and Browser is untouched.
- Uncheck Browser in the Layout dialog and press **Save Settings** → the setting stays `false` (the stale-draft path that would have reverted it).
- Reload with both off: neither rail returns, the Layers panel is unaffected.
- Re-enable each from Settings → the panel comes back collapsed on its rail and survives the next reload.
- No new console warnings (no `openRightPanel: no right panel registered`).

`tests/layout-panel-settings.test.ts` covers the defaults, the round-trip of a disabled panel, the back-compat fallback for settings saved before the keys existed, and rejection of non-boolean values. Full frontend suite (6040 passing), `npm run build`, and pre-commit are green.

## Summary by CodeRabbit

* **New Features**
  * Browser and Comments panel visibility preferences are saved and restored across app restarts.
  * Settings changes are applied when saved and remain synchronized with the active layout.
  * Settings displays the current panel visibility, and resetting the layout restores both panels to visible.

* **Bug Fixes**
  * Invalid or legacy visibility settings safely fall back to the default configuration.
  * Panel state is preserved when panels are collapsed, reopened, or removed during cleanup.

### DIFF
--- a/apps/geolibre-desktop/src/components/layout/SettingsDialog.tsx
+++ b/apps/geolibre-desktop/src/components/layout/SettingsDialog.tsx
@@ -12,12 +12,6 @@ import {
   type RuntimeEnvironmentVariable,
 } from "@geolibre/core";
 import {
-  closeRightPanel,
-  collapseRightPanel,
-  isRightPanelVisible,
-  openRightPanel,
-} from "@geolibre/plugins";
-import {
   Button,
   Dialog,
   DialogContent,
@@ -96,6 +90,7 @@ import { COMMENTS_PANEL_ID } from "../../hooks/useRegisterCommentsPanel";
 import { useRightPanelState } from "../../hooks/useRightPanels";
 import type { ThemeMode } from "../../hooks/useThemeMode";
 import { isTauri } from "../../lib/is-tauri";
+import { applyRightPanelVisibility } from "../../lib/persisted-right-panel";
 import { COORDINATE_FORMATS, normalizeCoordinateFormat } from "../../lib/coordinate-format";
 import { THEME_SCHEMES, normalizeHexColor, type ThemeScheme } from "../../lib/theme-schemes";
 import { IS_MAS_BUILD } from "../../lib/build-flags";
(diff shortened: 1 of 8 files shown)


giswqs:
Both review findings were real and are fixed in c0678c22. The underlying cause of each was the same: I had left the Browser/Comments rows in the Layout **dialog** applying live while the other four rows waited for Save, so they sat outside that dialog's draft/Save/Cancel contract.

Both rows are now draft-backed like their neighbours:

- **Reset then Cancel** (finding 1): `resetLayoutSettings` is back to touching only the draft, so nothing is persisted until Save and Cancel discards the whole reset, panel flags included. Verified: unchecking two rows, pressing Reset, then Cancel leaves stored `layout` byte-identical and the panels untouched.
- **Checkbox/draft divergence** (finding 2): the draft seeds those two keys from `isRightPanelVisible(...)` when the dialog opens rather than from the stored value. Closing a panel from its own header still shows an unchecked box, and Save now persists that unchecked value instead of writing the stale `true` back. Verified: close Browser from its header, open Settings (box is unchecked), Save, reload, and it stays closed.

Save applies the committed values to the registry, guarded on `isRightPanelVisible(id) === show`, so saving a dialog where those rows were untouched cannot collapse a panel the user had expanded. Verified with Comments expanded and an unrelated Layout row saved.

The Settings **dropdown** toggles are unchanged: they have no Save to wait for, so they move the panel and persist the preference in one step, matching the other live entries in that menu.

On the test-coverage note: agreed that the interactive logic is untested, and it is the existing convention here that these dialogs are covered through the normalize layer. The behaviors above were verified by driving the built app instead, in both dark and light themes. Full frontend suite (6040 passing), build, and pre-commit are green.

giswqs:
Good catch, and it is the right read: seeding the draft from the live registry closed the previous round's finding but only moved the seam. A panel closed from its own header was still session-only state that an unrelated Save then promoted to a permanent preference.

Fixed in 66280e0f by removing the session-only state rather than trying to guard the Save path. The new `apps/geolibre-desktop/src/lib/persisted-right-panel.ts` seeds the panel from its setting at registration and then subscribes to the registry, writing every later visibility change back. For these two panels closing is not a transient collapse: it removes the rail entry entirely and only Settings can restore it, so it is a preference however it was reached. With the setting and the registry always in step there is nothing left for a Save to silently promote, and the checkbox cannot disagree with either.

Two registry events are deliberately not mirrored, both called out in the module doc:

- the emit from registration itself, since the panel is legitimately not visible yet, so the subscription is attached after the seed;
- the emit from unregistering on unmount, which would otherwise persist `false` on every teardown. `unregisterRightPanel` removes the panel before it emits, so the `getRightPanel` guard catches it, and the disposer unsubscribes before disposing.

Being displaced by another panel is not a close (the registry keeps a displaced panel in `visibleIds`), so it correctly writes nothing.

On the coverage point, which was fair on both rounds: `tests/persisted-right-panel.test.ts` now covers the seed in both directions, the header close, the collapse and displacement non-events, the teardown guard, and the bail-out that keeps an untouched Save from collapsing an expanded panel. The module imports `@geolibre/plugins/right-panel-registry` rather than the package barrel so it stays a leaf the suite can load, per the CLAUDE.md guidance about not dragging the whole plugin registry into the coverage denominator.

Verified in the built app in both themes: header close persists and survives reload; the dropdown and dialog checkboxes agree; Reset then Cancel persists nothing; Reset then Save reopens and persists; expanding does not flip the setting and an unrelated Save leaves an expanded panel expanded. 6050 frontend tests passing, build and pre-commit green.

PR #1961, opened 2026-08-16
fix(map): allow grabbing the map during inertial pan

### LINKED ISSUE #1959: [Bug]: Grabbing the map during inertial pan aborts the new drag
### Before you start

- [x] This is a bug report, not a usage question.

### What happened?

After panning the map and releasing the mouse, MapLibre keeps the camera moving with inertial coasting. If I press and drag again **before that coasting has finished**, the map stops sliding and the new drag does not take over. I have to release and start a second drag to pan again.

This is not MapLibre's default gesture behavior (its own demos let you grab the map during inertia and continue the pan). It is also not an intentional GeoLibre UX choice: the store-as-camera-source-of-truth sync is intentional, but aborting an in-progress user drag is a side effect of that loop.

### Steps to reproduce

1. Run the web app (`npm run dev` → http://localhost:5173, or the published web build).
2. Click and drag the map, then release with some velocity so inertial coasting continues.
3. Before the map comes to rest, click and drag again.
4. The coasting stops, and the map does not follow the new drag until mouseup and another drag.

### Expected behavior

Grabbing the map while it is still coasting should stop inertia **and** continue as a new pan on the same pointer-down, the way MapLibre, OSM, and Google Maps do.

Web app from source (`npm run dev`), GeoLibre 2.6.0, MapLibre GL JS 6.3.0

Ubuntu 24.04 + Chromium-based browser

### Sample data

None — reproduces on the default basemap with no extra layers.

### Screenshots or logs

No console errors. The failure is interaction-only.

### Cause

`MapCanvas` writes the live camera into the Zustand store on every `moveend`, and a React effect then calls `applyView` → `map.jumpTo(...)` whenever `mapView` changes.

Interrupting inertia is itself a `moveend` (MapLibre `easeTo` is `stop()`ed by the new pointer). That store write re-enters `applyView` while `dragPan` is already active. MapLibre's `jumpTo` calls `Camera.stop()` **without** allowing gestures, which `reset()`s the drag handlers, so the new pan is aborted.

The same feedback loop is already documented for story-map playback (`jumpTo` cancels an in-flight chapter fly); user panning during inertia was not given an equivalent guard.

Relevant code:

- `packages/map/src/MapCanvas.tsx` — `moveend` → `setMapView`, effect → `applyView`
- `packages/map/src/map-controller.ts` — `applyView` always `jumpTo`
- MapLibre `Camera.jumpTo` / `_stop(allowGestures)` in v6.3.0

### Suggested fix

- Skip `jumpTo` when a camera gesture is active (`dragPan.isActive()` / `dragRotate.isActive()`).
- Skip a no-op `jumpTo` when the live camera already matches the store (even an identical `jumpTo` still `stop()`s handlers).
- Do not echo a map-originated `moveend` back through the `applyView` effect (the story presenter already skips this path for the same reason).

Happy to open a PR if that approach looks right.

mfkj8866:
## Summary
- Skip `applyView` → `jumpTo` while `dragPan` or `dragRotate` is active, so grabbing the map during inertial coasting does not abort the new drag (`jumpTo` calls `Camera.stop()` and resets handlers).
- Fixes #1959.

## Test plan
- [x] `tests/map-controller.test.ts`: no `jumpTo` while dragging or rotating; `jumpTo` still runs when idle
- [x] `npm run test:frontend` (6200 pass, 0 fail)
- [ ] Drag the map, release with inertia, grab again before it stops — the pan should continue

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented automatic map repositioning from interrupting active dragging or rotation.
  * Map view synchronization continues normally once user interactions are inactive.

* **Tests**
  * Added coverage for map behavior during active and inactive dragging and rotation.
  * Verified that requested map centers are applied when no interaction is active.

### DIFF
--- a/packages/map/src/map-controller.ts
+++ b/packages/map/src/map-controller.ts
@@ -1121,6 +1121,8 @@ export class MapController {
 
   applyView(view: MapViewState): void {
     if (!this.map) return;
+    // jumpTo stop()s drag handlers, so skip while the user is still panning.
+    if (this.map.dragPan.isActive() || this.map.dragRotate.isActive()) return;
     this.map.jumpTo(constrainMapView(view, this.mapPreferences, this.map));
   }
 
--- a/tests/map-controller.test.ts
+++ b/tests/map-controller.test.ts
@@ -180,6 +180,9 @@ function makeFakeMap(initialBasemapLayers: string[] = ["basemap-bg"]): {
     // A laid-out viewport, so the camera helpers that scale with the canvas
     // (the globe-safe fit ceiling) have a real size to work from.
     getCanvas: () => ({ clientWidth: 576, clientHeight: 648 }),
+    jumpTo: record("jumpTo"),
+    dragPan: { isActive: () => false },
+    dragRotate: { isActive: () => false },
     flyTo: record("flyTo"),
     fitBounds: record("fitBounds"),
     cameraForBounds: (...args: unknown[]) => {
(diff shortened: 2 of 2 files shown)


giswqs (approved):

PR #2065, opened 2026-08-22
feat(layers): show a tiled archive as a group of its source layers

### LINKED ISSUE #2062: [Feature]: Show a tiled archive's sublayers and toggle them individually
## What problem does this solve?

A PMTiles, MBTiles or vector-tiles archive usually holds several source layers. GeoLibre shows one
row for the whole archive: one name, one colour, one visibility toggle.

The Protomaps basemap is nine source layers — earth, water, roads, buildings, landcover, landuse,
places, pois, boundaries. Added to the map it is a single flat wash. Nothing lists what is in it,
nothing says which part is which, and the earth polygon that covers everything else cannot be
turned off.

Colours for those parts are already computed at add time and then discarded: the PMTiles control
assigns one per source layer, and only the first reaches the map.

## Proposed solution

Show an archive's source layers under it in the Layers panel — one row each, with the layer's name,
its colour, and its own visibility toggle. Hiding a part should stick when the project is saved and
reopened.

Colours should come from what the control already assigned, so an archive is legible as soon as it
is added rather than after manual restyling. Changing the layer's colour by hand should still win.

Applies to every archive kind that has source layers (PMTiles, MBTiles, vector tiles), not just
PMTiles.

## Sample data

`https://demo-bucket.protomaps.com/v4.pmtiles` — the Protomaps basemap, nine source layers
(`boundaries`, `buildings`, `earth`, `landcover`, `landuse`, `places`, `pois`, `roads`, `water`),
anonymous and CORS-open, so it adds straight from Add Data → PMTiles Layer. Fetch lists the nine;
after adding, the panel shows one row.

## Estimated scope

Medium: a broader feature that is feasible in a reasonable timeframe

## Alternatives considered

Add each source layer as its own layer in the panel. Fits the existing model, but a basemap archive
becomes nine rows, and those rows would share one underlying source — removing one must not break
the rest.

Track hidden source layers only, without touching styling. Smaller, but per-part symbology is the
obvious next step and would replace it.

## Anything else?

A follow-up worth keeping in view: some catalogs publish symbology for a *single* source layer split
by attribute — a fault map with 16 hazard classes, each its own colour and dash pattern. Whatever
shape holds "part of a layer" here should be able to describe that too.

clintonlunn:
Closes #2062

after-sublayers-group-per-source-layer

## What

A vector PMTiles archive is added as a folder named after it, with one layer per source layer
inside. The Protomaps basemap arrives as `v4` holding `water`, `roads`, `buildings`, `earth` and the
rest, each in the colour the control assigned it.

Nothing new in the panel: they are ordinary layers in an ordinary group, so visibility, opacity,
reordering, zoom-to, the Style panel and delete all work on them already.

A raster archive, or one with a single source layer, is added as one layer as before.

## Why this shape

The first cut gave the layer a list of parts with their own toggles — a second, parallel way to
express visibility and styling. The maintainer's suggestion on #2062 was to use layer groups, which
already carry collapse, group visibility ANDed with each child, group opacity multiplied into each
child, and nesting. That deleted more code than it added and gave per-source-layer styling for free.

## The part that needed care

The layers share one MapLibre source. `removeLayerFromMap` removed a layer's sources unconditionally,
so deleting one source layer would have pulled the source out from under its siblings — with a delete
button now on every row, that is one click away. It now takes the surviving layers and keeps a source
while anything still draws from it.

## Colours where an archive is still one layer

The STAC panel and the offline basemap extract build a single layer over all the source layers.
Those still paint each one in the colour the archive assigned (`assignedSourceLayerColor`), which
was already computed at add time and previously discarded after the first. A user restyling the
layer takes it back.

## Tests

`pmtiles-archive-layers.test.ts` covers the expansion and the refcount, including that a shared
source survives one sibling's removal and goes when the last one does. `layer-parts-every-path.test.ts`
pins the assigned colours through the vector-tiles and MBTiles sync paths.

6770/6771 pass, tsc clean.

## New Features
- PMTiles vector archives are split into source-layer overlays, while raster archives remain single overlays.
- Archive layers are grouped for easier management, with improved add, replace, and removal behavior.
- Existing styling and shared data sources are preserved during updates.
- PMTiles control selections and generated layers stay synchronized without duplicates.

## Bug Fixes
- Prevented shared sources and unrelated archive layers from being removed prematurely.
- Improved cleanup of empty archive groups and stale selections.

## Documentation
- Added guidance describing PMTiles archive layering behavior.


giswqs (approved):
Another great addition! Thank you for your contribution!

Image

PR #2126, opened 2026-08-26
fix(style-import): keep a stack whose class layer names no colour, per the spec default (#2125)

### LINKED ISSUE #2125: A class layer with no explicit color stops a whole style importing as rules
### What happened?

#1957 combines filtered same-type layers into rule-based symbology. One layer
without an explicit color disqualifies the whole set, so only the first layer is
imported — a real 16-class fault style arrives as 1 color for 16 classes.

```json
{"layers": [
  {"id": "a", "type": "line", "paint": {"line-color": "#e60000"},
   "filter": ["==", ["get", "class"], "a"]},
  {"id": "b", "type": "line", "paint": {"line-width": 2},
   "filter": ["==", ["get", "class"], "b"]}]}
```

`"line-color": null` behaves the same way.

### Expected behavior

A layer with no color takes the style spec's default (`#000000` for
`line-color`) and joins the rules. That is what MapLibre renders it as, so the
import matches what the publisher sees.

clintonlunn:
Closes #2125.

Basically, what's happening is: a style file says "draw these five classes in these
colours". Two of the five name no colour of their own, which is normal — MapLibre
draws those in the default, black. GeoLibre saw the two blanks, gave up on the whole
file, and painted all five classes in the first class's red.

**Before** — five classes, all red.

1-before-prod

**After** — the colours the file asked for, with the two blank classes in black.

2-after-branch

<details>
<summary>The style used in both screenshots</summary>

`residential` and `track` name no colour. Every other class does.

```json
{
  "version": 8,
  "layers": [
    { "id": "primary", "type": "line", "source-layer": "roads",
      "filter": ["==", ["get", "roadclass"], "primary"],
      "paint": {"line-color": "#e60000", "line-width": 3} },
    { "id": "secondary", "type": "line", "source-layer": "roads",
      "filter": ["==", ["get", "roadclass"], "secondary"],
      "paint": {"line-color": "#e69800", "line-width": 3} },
    { "id": "tertiary", "type": "line", "source-layer": "roads",
      "filter": ["==", ["get", "roadclass"], "tertiary"],
      "paint": {"line-color": "#4ce600", "line-width": 3} },
    { "id": "residential", "type": "line", "source-layer": "roads",
      "filter": ["==", ["get", "roadclass"], "residential"],
      "paint": {"line-width": 3} },
    { "id": "track", "type": "line", "source-layer": "roads",
      "filter": ["==", ["get", "roadclass"], "track"],
      "paint": {"line-width": 3} }
  ]
}
```

Applied to a generated GeoJSON of plain lines carrying a `roadclass` field.

</details>

---

A class layer that names no colour takes the style spec's default (`#000000`) and
joins the rules, instead of disqualifying the whole stack. An explicit `null` counts
as naming none — MapLibre reads it as unset and paints the same default.

Two cases still decline the stack, because importing them as a flat colour would be
wrong rather than incomplete:

- A colour present as something *else* — an expression, an interpolation.
- A class that names no colour but paints the feature another way. The spec encodes
  this as `line-color`'s `requires: [{ "!": "line-pattern" }]`; `line-gradient` and
  `fill-pattern` take over the same way. These would otherwise import an ArcGIS
  hatch class as solid black.

`SPEC_DEFAULT_COLOR` is hard-coded rather than read from the spec, because
`@geolibre/map` is published and does not depend on it. Both directions are guarded
by tests, per the mirror-constant convention in `docs/maintenance.md`.

Scope is colour only. Hidden classes (`layout.visibility: "none"`) import as enabled
and widen the layer's zoom range — that predates this change and reproduces on `main`
today for any stack whose classes all carry colours, so it is left alone here.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved MapLibre style imports for filtered line layers with omitted or `null` colors.
  - Preserves valid rule-based imports using the default black color instead of incorrectly flattening or rejecting styles.
  - Correctly avoids flattening styles when patterns, gradients, or data-driven color expressions override the default.

- **Documentation**
  - Added guidance for style color defaults and overriding paint properties.

- **Tests**
  - Added coverage for default colors, patterns, gradients, and data-driven color expressions.


giswqs (approved):

PR #2325, opened 2026-09-08
fix(plugins): allow clearing pins on uninstall and re-trusting blocked plugins (#2318)

### LINKED ISSUE #2318: [Bug]: A URL-installed plugin blocked by the SHA-256 integrity pin on its first load can never recover through the UI
### Before you start

### What happened?

When a remotely-installed plugin's bundle hash changes since it was last trusted, loadPluginUrlBundles correctly blocks it and shows "Plugin at '...' changed since you last trusted it and was not loaded." However, if this happens on the plugin's very first load attempt in the current page session (i.e. it never successfully registers), there is no way to recover through the UI:

Uninstalling and reinstalling does not clear the stale pin, because unloadRemovedUrlPlugins only drops a pin for URLs found in externallyLoadedPluginSources — a map only populated for plugins that successfully registered.
The "Update" button never appears, because isUpgradeable requires loadedVersions.get(entry.id) !== undefined — also only true for a plugin that previously registered.
Even if reached directly, reloadExternalUrlPlugin throws Cannot update plugin: no loaded version was found... Try reloading the app for the same reason — and reloading the app does not help, since the pin lives in localStorage and survives the reload.
The only recovery is manually editing localStorage["geolibre.pluginBundlePins"] in DevTools.

### Steps to reproduce

1.Install a plugin from a manifest URL (Manage Plugins → paste a manifest URL → Install), let it load successfully once.
2.Change the plugin's served bundle content (e.g. edit and rebuild its entry.js) without changing its manifest URL.
3.Reload the app. The plugin now shows "Failed" with the "changed since you last trusted it" message.
4.Try to recover: click the trash icon to uninstall, then reinstall the same URL — still fails with the same message.
5.Look for an "Update" button — none appears (Upgradeable (0)).

### Expected behavior

Either uninstalling a "Failed (changed)" plugin should always clear its pin (regardless of whether it ever successfully registered), or the "Update"/reload action should be reachable and effective for a plugin that has never successfully loaded, so a legitimate content change can be reviewed and re-trusted without editing localStorage manually.

Web app

_No response_

### Sample data

_No response_

### Screenshots or logs

_No response_

RohithPariki:
Closes #2318

### Summary of Changes

When a URL-installed plugin bundle changes after its initial trust, it is blocked from loading by its SHA-256 integrity pin (`plugin-integrity.ts`). However, when this occurred on the first load of a session (before the plugin ever registered in the plugin manager):
1. **Uninstalling left stale pins in `localStorage`**: `unloadRemovedUrlPlugins` only dropped pins for URLs matching registered entries in `externallyLoadedPluginSources`. Because the blocked plugin never registered, its pin remained persisted, preventing reinstallation even after removal.
2. **"Update" action was unavailable**: `isUpgradeable` required `loadedVersions.get(entry.id) !== undefined`, which was false for blocked plugins.
3. **`reloadExternalUrlPlugin` threw an error**: Even if invoked directly, it rejected with `Cannot update plugin: no loaded version was found...` when `existingId === null`.

This PR fixes the recovery path across all three layers:
- **Pin cleanup on removal**: `unloadRemovedUrlPlugins` now checks all pinned manifest URLs via `listPinnedPluginUrls()` and removes pins for any URL not present in the active `keep` set, guaranteeing stale pins are cleared on uninstall even if the plugin never registered.
- **Support reloading un-registered plugins**: `reloadExternalUrlPlugin` now handles `existingId === null` gracefully, fetching and registering the updated bundle and recording the new trusted pin.
- **UI recovery in `ManagePluginsDialog`**: `isUpgradeable` now returns true for installed plugins with active load issues (such as integrity pin blocks), allowing users to review and update/re-trust the plugin.
- **State notification**: `upgradeExternalPlugin` clears `externalPluginLoadIssues` on success and notifies listeners so error badges disappear immediately.
- **Unit tests**: Added regression tests in `tests/plugin-integrity.test.ts` verifying `listPinnedPluginUrls()` and pin removal on uninstallation.

### Testing Done

- Added unit tests in `tests/plugin-integrity.test.ts`:
  - Verified `listPinnedPluginUrls` tracks and returns all stored pins.
  - Verified `unloadRemovedUrlPlugins` drops pins for uninstalled URLs that failed initial load and never registered.
  - Verified all tests pass via `node --import tsx --test tests/plugin-integrity.test.ts`.
- Validated TypeScript and Vite production bundle (`npm run build -w geolibre-desktop`).
- Validated ESLint (`npx eslint apps/geolibre-desktop/src/... tests/plugin-integrity.test.ts`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Plugin load failures now take priority over “Update available” messaging, making the plugin’s actual status clearer.
  * Updating an affected external plugin now clears its previous failure state after a successful reload.
  * Fixed stale integrity verification data that could prevent removed or previously failed plugins from loading correctly.
  * External plugins that were not loaded successfully can now be registered during a reload.
  * Improved plugin handling when plugins are removed while loading.

* **Tests**
  * Added coverage for plugin integrity cleanup and verification-pin management.

### DIFF
--- a/apps/geolibre-desktop/src/components/layout/ManagePluginsDialog.tsx
+++ b/apps/geolibre-desktop/src/components/layout/ManagePluginsDialog.tsx
@@ -215,18 +215,21 @@ export function ManagePluginsDialog({
   // An update is available only when the registry version is strictly newer
   // than the loaded one (directional, not any mismatch). isNewerVersion orders
   // a pre-release below its release, so an rc user is offered the GA build.
+  // A plugin blocked by a load failure (such as an integrity pin mismatch, #2318)
+  // also offers the Update action so users can re-fetch and re-trust it.
   const isUpgradeable = useCallback(
     (entry: PluginRegistryEntry) => {
+      if (!isInstalled(entry)) return false;
+      if (externalLoadIssues.has(entry.manifestUrl)) return true;
       const loaded = loadedVersions.get(entry.id);
       const ownsLoadedPlugin = pluginManifestUrlsForIds([entry.id]).includes(entry.manifestUrl);
       return (
-        isInstalled(entry) &&
         ownsLoadedPlugin &&
         loaded !== undefined &&
         isNewerVersion(entry.version, loaded)
       );
     },
-    [isInstalled, loadedVersions],
+    [isInstalled, loadedVersions, externalLoadIssues],
   );
 
   // True when the entry is in settings (so the badge reads "Installed") but the
@@ -637,7 +640,7 @@ export function ManagePluginsDialog({
                                 {category}
                               </span>
                             ))}
-                            {updateAvailable ? (
+                            {updateAvailable && !loadIssue ? (
                               <span className="rounded-full border border-amber-500/40 bg-amber-500/10 px-1.5 py-0.5 text-amber-600 dark:text-amber-400">
                                 {t("managePlugins.updateAvailable")}
                               </span>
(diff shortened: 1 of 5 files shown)


giswqs:
Superseded by #2322

PR #2362, opened 2026-09-12
fix(zarr): keep Zarr layers visible past zoom 12 on Mesa GPUs

### LINKED ISSUE #2357: [Bug]: Zarr layer stops rendering above zoom ~12
### Before you start

### What happened?

Zarr layer stops rendering above zoom ~12 — reproducible with the bundled
NOAA sample dataset

## Environment

- GeoLibre Desktop 2.9.0 (also reproduced on 2.7.0)
- `@carbonplan/zarr-layer` ^0.9.0
- Linux, WebKit (Tauri)

## What happens

A Zarr layer renders correctly up to map zoom ~11.9. At zoom 12 and above
the layer disappears completely — no error in the UI, the layer stays
listed and visible in the Layers panel.

## Why this is not a data problem

Reproduced with **two datasets of completely different size**:

| dataset | extent | behaviour |
|---|---|---|
| bundled NOAA sample ("Load sample data…") | global (~360°) | stops at ~zoom 12 |
| own dataset, single farm field | 0.023° (0.0065% of globe width) | stops at ~zoom 12 |

That the cutoff lands at the same zoom for datasets differing by a factor
of ~15,000 in extent argues against an extent-dependent cause and for a
fixed threshold somewhere in the render path.

## What was ruled out on our own dataset

Before testing the bundled sample we varied nearly everything:

- Zarr **v2 and v3** (v3 is required by the app; v2 gives
  `Not found: v3 array or group`)
- **UTM (EPSG:32613) and WGS84** coordinates
- **compressed (zstd) and uncompressed** (`bytes` codec only)
- **no pyramid**, OME-style `multiscales:[{datasets:[{path:"0"},…]}]`,
  and `{layout:[{asset,"spatial:shape","spatial:transform"}]}`
- a pyramid spanning **80 m down to 2.5 m** (6 levels), so a level finer
  than the native resolution exists for high zoom

All combinations behave identically: fine up to ~12, gone above.

Per-level transforms were verified to be consistent (every level covers
the same bounding box).

## Code reading (may be wrong — offered as a starting point)

`selectLevelForZoom` in `region-math.ts` has no upper cutoff: when no
level qualifies it falls back to the finest available level
(`return levelResolutions[levelResolutions.length - 1].index`). So level
selection alone should not produce a hard stop.

Two candidates we could not distinguish from the outside:

1. `getVisibleRegions` returns an **empty array** when no proj4
   transformer is set up (region-math.ts, ~lines 60-67) — a silent blank
   path. Should not apply to a stock WGS84 sample, but it is the one
   place that produces exactly this symptom without an error.
2. `MAX_CACHED_REGIONS = 128` in `region-cache.ts` versus
   `updateVisibleRegions`, which fetches every visible region without
   checking that count against the cache size. Visible regions grow with
   zoom; if the count crosses 128, regions could be evicted before upload
   and never stabilise.

(2) would explain the extent-independence: what grows with zoom is the
number of visible regions, not the geographic extent.

### Steps to reproduce

1. Load zarr dataset
2. zoom in onto a level of 12 and higher 

### Expected behavior

Map still shows something on all zoom levels. in Precision Farming i have 1x1m rasters and still want to see the 10x10m raster in the background 

Desktop App

Ubuntu 24.04.5 LTS

### Sample data

Inegrated NOAA Dataset

### Screenshots or logs

_No response_

giswqs:
## Summary

Fixes #2357. A Zarr layer with a CRS disappeared on laptops with integrated graphics (Intel and AMD, Mesa drivers) as soon as the map left the globe transition: zoom 12 and above with the globe projection, and every zoom in plain Mercator. NVIDIA never showed it, which is why it did not reproduce until the run was forced onto Mesa.

## Root cause

`@carbonplan/zarr-layer` compiles its flat source-projected shader at that point. There `gl_Position` comes from `u_anchor_clip + deltaClip`, so `shift_x`, `shift_y` and `u_worldXOffset` only feed the `v_mercatorPos` varying, which the fragment shader never reads. Mesa eliminates that chain at link time and reports the three uniforms inactive; the renderer's strict `mustGetUniformLocation` then threw `Failed to get uniform location for shift_x` on every frame, the program was never cached, and MapLibre's frame loop swallowed the exception, so the layer silently stopped drawing with nothing in the UI. NVIDIA keeps the uniforms active.

Confirmed by dumping active uniforms in WebKitGTK on llvmpipe: the ECEF variant used below zoom 12 keeps all three, the flat variant compiled at zoom 12 loses them.

## Change

- Carry the upstream fix (carbonplan/zarr-layer#91) as a patch-package patch until a release ships it: look those three uniforms up with the non-throwing `gl.getUniformLocation`, as the renderer already does for `matrix` and `u_eye_matrix`. A null location is a no-op for `gl.uniform1f`.
- `tests/zarr-layer-patch.test.ts` pins the installed bundle so a dependency bump that silently drops the patch fails the suite (it fails against the pristine 0.9.0 bundle).
- Refresh the comment in `scripts/apply-dependency-patches.mjs` now that two app dependencies are patched.

Once the upstream release lands, bump `@carbonplan/zarr-layer` and delete the patch and the test.

## Verification

- Mesa llvmpipe in WebKitGTK (the Tauri engine), NOAA sample: before the patch the layer vanished at zoom 12.5 with `Failed to get uniform location for shift_x` thrown each frame; after it the layer renders at zoom 12.5 and 14 in globe and at 11.5 and 13 in Mercator with zero errors.
- NVIDIA GL in Chrome against the patched dev server: unchanged, renders at every zoom in both projections, also under mouse-wheel zoom at device scale factor 2.
- Dark theme (WebKit harness) and light theme (Chrome) covered by the two runs above.
- `npm run build`, the new test, and `pre-commit run --files` on the changed files pass.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved compatibility for flat, source-projected rendering on GPU drivers that omit certain inactive shader uniforms.
  - Prevented shader initialization failures in affected rendering scenarios, helping layers display reliably across more hardware configurations.
  - Preserved required uniform handling for opacity and other active rendering properties.

### DIFF
--- a/patches/@carbonplan+zarr-layer+0.9.0.patch
+++ b/patches/@carbonplan+zarr-layer+0.9.0.patch
@@ -1,0 +1,25 @@
+diff --git a/node_modules/@carbonplan/zarr-layer/dist/index.js b/node_modules/@carbonplan/zarr-layer/dist/index.js
+index 9fc526c8..213a1f8b 100644
+--- a/node_modules/@carbonplan/zarr-layer/dist/index.js
++++ b/node_modules/@carbonplan/zarr-layer/dist/index.js
+@@ -2497,9 +2497,17 @@ function createShaderProgram(gl, options) {
+     scaleLoc: mustGetUniformLocation(gl, program, "scale"),
+     scaleXLoc: mustGetUniformLocation(gl, program, "scale_x"),
+     scaleYLoc: mustGetUniformLocation(gl, program, "scale_y"),
+-    shiftXLoc: mustGetUniformLocation(gl, program, "shift_x"),
+-    shiftYLoc: mustGetUniformLocation(gl, program, "shift_y"),
+-    worldXOffsetLoc: mustGetUniformLocation(gl, program, "u_worldXOffset"),
++    // In the source-projected flat (wgs84) shader gl_Position comes from
++    // u_anchor_clip + deltaClip, so shift_x/shift_y/u_worldXOffset only feed
++    // the v_mercatorPos varying, which the fragment shader never reads. Mesa
++    // (Intel/AMD/llvmpipe) eliminates that chain at link time and reports the
++    // uniforms inactive, which made mustGetUniformLocation throw on every
++    // frame from the moment the map left the globe transition, so the layer
++    // silently vanished at zoom >= 12 on those GPUs. A null location is a
++    // no-op for gl.uniform1f, so look these up without throwing.
++    shiftXLoc: gl.getUniformLocation(program, "shift_x"),
++    shiftYLoc: gl.getUniformLocation(program, "shift_y"),
++    worldXOffsetLoc: gl.getUniformLocation(program, "u_worldXOffset"),
+     // MapLibre modes use projectTile instead of matrix. The Mapbox flat
+     // source-projected (wgs84) shader uses u_eye_matrix, not matrix, so the
+     // compiler drops `matrix` there too — use the non-throwing lookup. All
--- a/scripts/apply-dependency-patches.mjs
+++ b/scripts/apply-dependency-patches.mjs
@@ -2,9 +2,10 @@ import { existsSync } from "node:fs";
 import { spawnSync } from "node:child_process";
 
 // Workspace-scoped production installs, such as the collaboration worker
-// image, do not install GeoLibre's raster dependencies. There is nothing to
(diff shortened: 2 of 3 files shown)


giswqs:
Layer now renders at zoom level > 12

image

PR #2370, opened 2026-09-13
feat(field-collection): support multiple photo attachments

kongdd:
## Summary

Support multiple photo attachments per Field Collection observation.

- Select multiple images, append more, and remove individual previews before saving.
- Store ordered image data URLs in `geolibre_photos` and filenames in `geolibre_photo_names`.
- Keep the first image in the existing `photo` property so current single-photo popups and exports remain compatible.
- Preserve the existing per-photo size checks and capture-context cancellation behavior; reserve the new property keys against custom form fields.

## Scope

This is the first, narrowly scoped Field Collection change. It does **not** add map-click record dialogs, project-folder access, new plugins, local filesystem mounts, dependencies, or build changes. Existing single-photo viewing behavior is unchanged.

## Verification

- Unit tests cover serialization, first-photo compatibility, empty/unnamed attachments, and reserved keys.
- Playwright covers multi-file selection, appending, individual removal, rejection of oversized batches, and the downloaded project's attachment properties.

Validation completed in an isolated worktree based on upstream `main`:

- `npm run test:frontend`: **8118 passed, 1 skipped, 0 failed**.
- `pre-commit run --files ...`: passed all applicable hooks, including formatting, ESLint and the production build.
- `e2e/field-collection-photos.spec.ts`: **passed** against that production build (using a preview-only local Playwright config).

The PR touches five files: the capture dialog, its pure serialization helper/tests, one browser spec, and the feature documentation.

## Summary by CodeRabbit

* **New Features**
  * Field Collection supports attaching multiple photos to each observation.
  * Add, preview, and remove individual photos before saving.
  * Photo names and image data are preserved when saving and exporting projects.
  * Existing single-photo displays and exports remain supported through the first attached image.

* **Bug Fixes**
  * Oversized or unreadable photos are rejected with clear error messages.
  * Saving is prevented until photo processing is complete.


kongdd:
Addressed the review in a focused follow-up:

- Track pending photo batches (including overlapping selections), disable/guard Save until they settle, and invalidate the capture after a successful save.
- Give each remove button a distinct accessible name including its position and filename; rename the `filter` index parameter.
- Clarify that the size cap applies per photo, not per observation/session. No new arbitrary attachment limit is introduced.
- Extend the browser regression to hold FileReader completions deterministically, exercise overlapping batches and read errors, and verify saved attachments. The new test fails on the previous build because Save remains enabled during a read.

Validation: frontend suite **8118 passed / 1 skipped**; scoped pre-commit hooks (including production build) passed; the extended Playwright regression passed against the rebuilt app.

Retained `Promise.all`: any rejected batch is discarded atomically and reports a valid failure reason. Deterministic priority between different read failures does not require expanding this PR to `allSettled`.

giswqs (approved):

Issue #1339, opened 2026-07-19
[Feature]: Add ability to copy/paste styles between layers


spatialthoughts:
### Before you start

- [x] I searched the [existing issues](https://github.com/opengeos/GeoLibre/issues) and did not find a duplicate request.

### What problem does this solve?

When trying to visualize multiple layers, it is useful to visualize them with the same parameters.

For example, I am loading 4 rasters with different masks and they all load with different default visualizations. I need to go and adjust each so they can be comparable

<img width="1150" height="816" alt="Image" src="https://github.com/user-attachments/assets/91e3618a-a9b4-4f2b-a1f5-ca670889e0aa" />

### Proposed solution

QGIS allows copy/pasting layer styles between vector as well as raster layers. A similar functionality will be useful

### Estimated scope

Small: a minor tweak, small UI change, or text adjustment

giswqs:
Good suggestion. The feature request has been implemented in #1342. 

<img width="815" height="517" alt="Image" src="https://github.com/user-attachments/assets/06cf1207-29e7-4df5-a647-7c90dd466205" />

PR #2385, opened 2026-09-14
feat(style-import): paste a style from the layer menu and the Style panel

### LINKED ISSUE #2384: [Feature]: Paste a style instead of opening a file
### Before you start

- [X] I searched the [existing issues](https://github.com/opengeos/GeoLibre/issues) and did not find a duplicate request.

### What problem does this solve?

A style import needs a file on disk. A style is often already in the clipboard. A user copies one
out of QGIS, a gist, or a data catalog page. GeoLibre cannot read it without a save to disk first.

The existing entry point is also hard to find. It sits under layer `⋯`, then Styles, then Import
style. A user who works in the Style panel must leave the panel to reach it.

### Proposed solution

Add a paste dialog. It holds a textarea, a Cancel button, and an Import button. Import stays
disabled until the textarea holds text.

Reach it from two places:

- The layer actions menu, next to the file import.
- The Style panel header, next to the Style Manager button.

A style behaves the same through either entry point. All four formats work through either entry
point: a GeoLibre style URL, Mapbox GL JSON, OGC SLD, and QGIS QML. The format is read from the text
and not from a file extension.

Errors stay in the dialog. They do not become a layer row note. The user is looking at the dialog.

The file import keeps its current behaviour and its current notes.

This needs the style reading to move out of `LayerPanel` first. It sits there as a closure, tangled
with the panel status notes, and no other caller can reach it. That move has no user-visible effect.

### Estimated scope

Medium: a broader feature that is feasible in a reasonable timeframe

### Alternatives considered

Read the clipboard directly with `navigator.clipboard.readText()`. That asks for permission in the
browser build. Firefox does not support it for web pages. A paste box works everywhere. It costs one
extra click.

clintonlunn:
## What this adds

A dialog that imports a style from pasted text. It reaches the same four formats the file import
reads. Two entry points open it: the layer actions menu and the Style panel header.

Closes #2384

## Commit 1: move the style reading into `@geolibre/map`

The format detection, the parsing, and the patch building sat inside `LayerPanel` as a closure. They
now live in `packages/map/src/style-import.ts` behind a `./style-import` subpath, so
`packages/plugins` can reach them.

```ts
importStyleText(text): { ok: true; apply(base): LayerStyle; warnings: string[] }
                     | { ok: false; reason: "invalid" | "no-match"; warnings: string[] }
```

The function takes text and returns a patch. Each caller renders its own status. `apply` does not
mutate its argument, because the caller passes the live store object.

`isQmlStyleXml` moved with it. `apps/geolibre-desktop/src/lib/style-format.ts` is deleted.
`StyleManagerPanel` imports from the new home.

A subpath rather than the package index: the index reaches the MapLibre stylesheet, which
`node --test` cannot load. The three parsers import only `@geolibre/core`, `fast-xml-parser`, and
sibling modules.

The file import is untouched. The picker call stays inside the `try` block that turns a rejection
into a layer row error. The pending status timer still clears when the user selects the menu item.

## Commit 2: the dialog and its two entry points

`PasteStyleDialog` is one shared component, so the try/catch, the reset, and the error rendering
exist once. It guards both `importStyleText` and the caller's `onApply`.

Both entry points gate on `isStyleLibraryTargetLayer`. The Style panel button sits inside the
existing gate rather than beside it.

State resets on every open, so the box is empty whatever closed it. The Style panel also closes the
dialog when the selected layer changes. The Layers panel holds a layer id rather than a boolean. A
paste is therefore bound to the layer it started from.

### Strings

The new keys are an `importStyleFromText*` family and deliberately not `pasteStyle*`. That name
belongs to the copy-between-layers feature (#1339) and has three siblings. `JSON.parse` drops a
duplicate key. The new label then never renders. The menu then shows two identical items one level
apart. A duplicate-key scan over `en.json` comes back clean.

`layers.importStyle` is reworded to "Import style from file (…)" so the pair reads as a pair.
`layers.importStyleInvalid` no longer says "file". The wording still serves the file door.

## A pasted QML fragment

`isQmlStyleXml` accepts a bare `<renderer-v2>` root, and `parseQml` read only a `<qgis>` root. A
fragment copied out of a `.qml` therefore routed to the QML parser and came back as `no-match`. The
file import never reached this, because a whole `.qml` carries the wrapper. The paste box does reach
it, so `parseQml` now reads the root itself when the wrapper is absent.

The detection had to widen with it. `parseQml` reads a root `<labeling>` too, and `isQmlStyleXml`
did not claim that root, so a bare labeling fragment went to the SLD parser. Both roots are now
detected, and a test asserts the two lists agree.

## What a failed import calls the thing it could not read

The parsers said "file" in their own warnings, and the paste box shows those warnings verbatim. A
paste has no file. All five now read "That", matching the reworded `layers.importStyleInvalid`. A
test asserts no failure message says "file" again.

## Docs

Three pages quoted `layers.importStyle` verbatim, so the rewording made them wrong. They now match
`en.json`, and the Layers page describes the paste box beside the file import.

- `docs/index.md`
- `docs/user-guide/layers.md`
- `docs/user-guide/embedding.md`

## Three places list a subpath export

Nothing checks that they agree. This PR updates all three, and a package build confirms them:

1. `exports` in `packages/map/package.json`
2. `publishConfig.exports`, hoisted at publish by `scripts/prepare-npm-package.mjs`
3. the `build` script's `tsdown` entry list

Miss 2 or 3 and the repository is fine while the published package is broken.

## Tests

- `tests/style-import.test.ts`: 28 tests over the format detection, the patch contract, and all
  three error shapes. One reads a bare `<renderer-v2>` fragment and a wrapped one, and asserts the
  wrapper is their only difference. Three more assert that no failure message says "file".
- `e2e/paste-style.spec.ts`: one spec per entry point. Each covers the disabled-until-text button,
  an error keeping the pasted text, and a successful import that reports the parser's warning.

That last assertion is the one that matters. "Style imported." alone passes whether or not the app
appends warnings. Warnings that silently stop appearing are the regression this reporting prevents.

## Gates

Run at `4906f9a0`.

- `npm run lint`: 0 errors, 40 pre-existing warnings
- `npm run typecheck`: exit 0
- `npm run test:frontend:coverage`: 8240 pass, 0 fail, 1 skip, exit 0, 87.08%
- `npm run i18n:tools:check`: exit 0
- `npx playwright test e2e/paste-style.spec.ts`: 2 passed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_013sLZChsa1KL9VY2Jo3HHQh

## Summary by CodeRabbit

- **New Features**
  - Import styles by pasting text directly into the Layers and Style panels.
  - Supports Mapbox GL JSON, OGC SLD, and QGIS QML formats, including QML fragments.
  - Displays parser warnings and clear success or error messages during import.
  - Added clearer file-import labeling and guidance for text-based imports.

- **Bug Fixes**
  - Preserves pasted content when validation fails.
  - Prevents applying styles to unavailable layers.
  - Ensures imported styles apply to the latest layer state.
  - Disables Apply until text is entered and resets dialog state when reopened.


clintonlunn:
@coderabbitai review

clintonlunn:
@coderabbitai full review

giswqs (approved):

PR #2386, opened 2026-09-14
feat: add getLayers/onLayersChanged to GeoLibreAppAPI

### LINKED ISSUE #2383: [Feature]: expose a layers changed subscription on GeoLibreAppAPI, alongside the existing onBasemapChange
### Before you start

- [x] I searched the [existing issues](https://github.com/opengeos/GeoLibre/issues) and did not find a duplicate request.

### What problem does this solve?

`GeoLibreAppAPI` already lets a plugin subscribe to basemap changes
(`getActiveBasemap()` / `onBasemapChange()` in `packages/plugins/src/types.ts`,
implemented via `useAppStore.subscribe` in
`apps/geolibre-desktop/src/hooks/usePlugins.ts`), but there's no equivalent
for the project's **layers**.

A plugin can add a layer (`addGeoJsonLayer`, `registerExternalNativeLayer`,
etc.) and remove one it owns (`removeLayer?`), but it has no way to find out
that a layer — including one it added itself — was removed by the user
through some other path (most commonly: clicking the trash icon on it in the
Layers sidepanel).

A plugin bundled inside the host can work around this today by importing
`useAppStore` directly (several built-in plugins already do this for
basemap changes, e.g. `maplibre-swipe.ts`, `maplibre-geohash.ts`,
`maplibre-s2.ts`), but an externally-loaded plugin has no access to the
store at all — `GeoLibreAppAPI` is its only surface, and that surface has no
way to observe the layer list changing.

Motivating case: we maintain an external plugin with a catalog panel — each
row has an "Add to Map" button that becomes a disabled "Added" label once
clicked. If the user later removes that layer from the Layers sidepanel, the
row has no way to notice, so its button stays stuck on "Added" until the
whole plugin is deactivated/reactivated or the page is reloaded.

### Proposed solution

Mirroring the existing basemap pair as closely as possible:

    /** Current layer ids in the project, in their current order. */
    getLayers: () => string[];

    /**
     * Subscribe to the project's layer ids. `callback` fires every time a
     * layer is added, removed, or reordered anywhere in the app (including
     * the user removing one from the Layers sidepanel, or another plugin
     * adding one). Returns an unsubscribe function.
     */
    onLayersChanged: (callback: (layerIds: string[]) => void) => () => void;

A plausible host implementation, following the exact same shape as
`onBasemapChange`:

    getLayers: () => useAppStore.getState().layers.map((l) => l.id),
    onLayersChanged: (callback: (layerIds: string[]) => void) =>
      useAppStore.subscribe((state, prev) => {
        if (state.layers !== prev.layers) {
          callback(state.layers.map((l) => l.id));
        }
      }),

This is additive only — a new method on `GeoLibreAppAPI`, implemented once
in `createAppAPI()`. Happy to adjust the exact shape/naming, and happy to
submit a PR if a maintainer can confirm the preferred approach first.

### Estimated scope

Small: a minor tweak, small UI change, or text adjustment

### Alternatives considered

- DOM polling: an external plugin could periodically scan the Layers
  sidepanel's rendered DOM for whether its tracked layer id/name is still
  present. Works in principle but is fragile (breaks silently if the
  sidepanel's markup changes) and wasteful compared to reacting to a real
  event.
- Bundling the plugin into the host instead of loading it externally:
  defeats the point of the external-plugin mechanism, and doesn't help
  other third-party plugin authors hitting the same gap.

Phichak-Kamin:
## Summary

Plugins can already subscribe to basemap changes via `onBasemapChange` /
`getActiveBasemap`, but had no equivalent for the layer list itself. This adds
the mirrored API for layers:

- `getLayers?: () => string[]` — the project's current layer ids, in order.
- `onLayersChanged?: (callback: (layerIds: string[]) => void) => () => void` —
  subscribes to the layer list; `callback` fires whenever a layer is added,
  removed, or reordered anywhere in the app (including the user removing one
  from the Layers panel, or another plugin adding one). Returns an
  unsubscribe function.

Both are typed optional, matching the current convention for newer
`GeoLibreAppAPI` methods (`addLayerGroup`, `getViewBounds`, `activatePlugin`,
etc.), unlike the older non-optional `getActiveBasemap`/`onBasemapChange` pair
this mirrors.

Implementation in `createAppAPI()` mirrors `onBasemapChange`'s own pattern:
subscribing to the Zustand store and diffing `state.layers` by reference.

Closes #2383

## Test plan

- [x] `npm run lint` — 0 errors (pre-existing warnings only, none on touched files)
- [x] `npm run build` (typecheck) — clean
- [x] `npm run test:frontend` — 8212 passed, 0 failed
- [x] `npx oxfmt --check` on both changed files — already correctly formatted

## Summary by CodeRabbit

- **New Features**
  - Plugin integrations can retrieve the project’s current layer IDs in order.
  - Plugins can subscribe to layer additions, removals, and reordering, with support for unsubscribing from updates.


giswqs (approved):

PR #2408, opened 2026-09-15
feat(map): host the layer control on the Mapbox renderer

giswqs:
## Summary

The on-map layer control (`maplibre-gl-layer-control`) now works on the Mapbox rendering engine, not just MapLibre.

The upstream control has no runtime imports and only uses the shared style API (`getLayer`, `getStyle`, paint/layout getters and setters, `moveLayer`, `fitBounds`, a `styledata` listener) plus `IControl`, all of which Mapbox GL JS implements. What pinned it to MapLibre was GeoLibre's own glue inside `MapController`. This PR extracts that glue into an engine-neutral `LayerControlHost` (`packages/map/src/layer-control-host.ts`) and has both `MapController` and `MapboxEngine` drive it through a small adapter. No upstream change was needed.

## What changed

- **New `layer-control-host.ts`**: the store-backed custom layer adapter, the structural rebuild signature, in-place mirroring of store visibility/opacity/names, the style-editor round trip, and the corner-order restore. `restoreControlOrder` and `layerControlPaintToStyle` move here (re-exported from `map-controller.ts`, so existing imports keep working).
- **`MapController`**: about 300 lines replaced by thin wrappers over the host. Behavior is unchanged except one improvement below.
- **`MapboxEngine`**: mounts the control on `style.load`, refreshes it after `syncLayers`, mirrors basemap state, drops it before a style swap, and answers the `layer-control` built-in control API (visible / position). Native layer ids come from the engine's compiled layer plans. A new constructor option lets split/grid panes opt out, matching `SecondaryMapCanvas`.
- **Basemap seeding**: the control learns which layers are basemap by fetching `basemapStyleUrl`. `mapbox://` styles cannot be fetched, and its fallback ("whatever was on the map at mount is basemap") sweeps up GeoLibre's own native layers, so the Background toggle would hide project layers. The host now seeds the control's basemap-id set from the engine's `getBasemapStyleLayerIds()` plus a sentinel (Mapbox Standard keeps its basemap in style imports and has zero root layers). Mapbox always uses this; MapLibre uses it for blank and `geolibre://` sentinel basemaps, where it previously had the same fallback.
- Mounting is guarded against `getStyle()` throwing while a style is still loading, on both engines.
- Docs: `docs/mapbox-renderer.md` supported paths, `docs/architecture.md` data-flow step.

## Verification

- `npm run test:frontend`: 8375 pass, 0 fail. New tests in `tests/mapbox-engine.test.ts` cover mount timing and pane opt-out, the built-in control API, rebuild-only-on-structural-change, style-swap remount, and (against a linkedom DOM) that the panel lists store layers, mirrors store state in place, and that hiding Background leaves the layer's native style layers alone.
- Browser check with a real Mapbox token (dev server, Playwright): loaded a project with `primaryRenderer: "mapbox"` and two inline GeoJSON layers. The control appears in the Mapbox map's top-right cluster, the panel lists both layers plus Background, unchecking a layer sets the store and the Mapbox layout visibility to `none`, the opacity slider updates the store and `circle-opacity`, and unchecking Background sets `basemapVisible` false (Standard fades to blank) while the project layers stay visible. Dark mode matches the MapLibre look. Switched back to MapLibre in the same session: control present with the same rows, and on a blank basemap the Background toggle hides only the blank background layer.
- `pre-commit run --files ...` clean (includes the full `npm build`).

## Notes

- Cesium is unaffected: it has no style document and continues to report `false` for the `layer-control` built-in.
- The layer control panel is light in dark mode on both engines; that is pre-existing and not changed here.

## Summary by CodeRabbit

* **New Features**
  * Added on-map layer controls for Mapbox maps, including visibility, opacity, ordering, and style editing.
  * Layer control changes stay synchronized with map and application state.
  * Controls refresh correctly when map styles change.
  * Basemap layers and visibility are reflected in the controls.
  * Secondary split-map panes hide duplicate controls to prevent conflicting updates.
  * Layer controls remain available when switching between supported map engines.

* **Documentation**
  * Updated architecture and Mapbox renderer documentation to describe layer-control behavior and style handling.


giswqs:
Follow-up commit: the Layer Control toggle lives under **Plugins → Layer Control** (a built-in plugin), and that plugin declared no `engines`, so the plugin manager treated it as MapLibre-only. On a swap to Mapbox it was deactivated, its `deactivate()` removed the control the engine had just mounted, and the menu entry greyed out. The plugin now declares `["maplibre", "mapbox"]`; verified in the browser that the entry is enabled under Mapbox and that hide/show and corner changes reach the Mapbox map, with a renderer test asserting the declaration.

giswqs:
Third commit: on Mapbox Standard the layer control opened empty for a project with no data layers. The upstream control only adds its Background row when the root style has basemap layers, and Standard keeps its basemap in style imports. The host now adds the Background row itself after mount when the control did not, and the toggle works as before through the store. Verified in the browser: fresh Mapbox map shows only Background, unchecking it fades Standard and the left Layers panel mirrors it, adding a layer shows the layer row above Background, removing it returns to Background alone.

PR #2453, opened 2026-09-17
fix(swipe): re-evaluate side visibility on styledata

### LINKED ISSUE #2434: Mapbox: a basemap change stops Layer Swipe applying its sides
Found while fixing #2430, and distinct from it. #2430 was the comparison pane being orphaned; with that fixed the control survives a basemap change and keeps its state, but the sides stop reaching the map.

## What happens

Layer Swipe active on `primaryRenderer: "mapbox"`, one project layer put on the **right** side only, then a style basemap picked from the Basemaps panel. Reading the `visibility` of that layer's style layers on each map:

| | main map | comparison pane |
| --- | --- | --- |
| before any change | `none` | `visible` |
| after 1 basemap change | `none` | **absent** |
| after 2 | absent | absent |
| after 3 | **visible** | absent |

The control's own state is intact throughout — `getState().rightLayers` still holds every `geolibre-mapbox-<id>-…` id — so this is the assignment not being re-applied, not a lost selection.

**The same sequence on MapLibre keeps `main: none` / `pane: visible` across every change**, with identical plugin code.

## Reproduce

1. View → Rendering engine → Mapbox.
2. Plugins → Layer Swipe → Activate.
3. In the panel, clear a layer from LEFT and check it under RIGHT. It disappears from the main map and draws on the clipped pane — correct.
4. Layers panel → Basemaps → pick **OpenFreeMap Positron** (a `STYLE` entry; a raster basemap is added as a layer and never reaches `setStyle`).
5. The layer is gone from the comparison pane. Repeat twice more and it comes back on the main map as well, i.e. it is no longer swiped at all.

## Where to look

The control is rebuilt on the new style's `style.load`, before the engine has re-added the project layers, so the `_updateLayerVisibility` inside `onAdd` has nothing to assign — the ids it holds match nothing yet. After that:

- `SwipeControl._styleDataHandler` (the persistent `styledata` listener) only runs `_applyDefaultSelectionIfPending` and `_refreshLayerList`; it never re-runs `_updateLayerVisibility`, which is also the only caller of `_syncLayersToComparisonMap` — the step that copies a right-side layer onto the pane. That is why the pane loses it first and permanently.
- `resolveSwipeProjectLayerIds` (`packages/plugins/src/plugins/maplibre-swipe.ts`) does re-run on `styledata` and would re-apply through `setLeftLayers`/`setRightLayers`, but only `if (left.changed)` / `if (right.changed)`. The ids are unchanged across a basemap change, so it no-ops — on MapLibre something else evidently lands the assignment, which is worth pinning down before choosing a fix.

Likely fixable upstream in `maplibre-gl-swipe` (re-assert visibility when the style changes under the control) rather than in GeoLibre, as with #2430.

## Coverage

`e2e/mapbox-swipe.spec.ts` ("keeps one comparison pane across basemap changes") asserts the carried-over sides but deliberately not what each map paints, precisely because of this. Tighten it to the visibility assertions when this is fixed.

RohithPariki:
Fixes #2434

## Root cause
When switching basemaps on Mapbox, `maplibre-gl-swipe` did not re-evaluate layer visibility after the new style loaded. The swipe panel stopped applying its sides correctly because the map engine resets visibility during a style change, but the control hadn't restored its internal state.

## Fix
- Patched `maplibre-gl-swipe` to call `_updateLayerVisibility()` inside its `styledata` handler so that side selections are re-applied immediately after a style change.
- Generated `patches/maplibre-gl-swipe+0.11.2.patch` via patch-package.

## Verification
- When switching basemaps with Layer Swipe active on the Mapbox renderer, the side selections are restored automatically.


giswqs:
Already fixed in #2435

PR #2455, opened 2026-09-17
fix(desktop): remove duckdb-wasm workarounds

### LINKED ISSUE #961: DuckDB-WASM read_parquet "stoi: no conversion" on GeoParquet — load-order workarounds tracking (upstream duckdb-wasm#2199)
### Summary

GeoLibre reads Parquet/GeoParquet client-side with `@duckdb/duckdb-wasm` (currently pinned `1.33.1-dev45.0`). On that build, `read_parquet` throws `Invalid Error: stoi: no conversion` while parsing GeoParquet metadata. The same files load fine in the DuckDB CLI, so this is a WASM-specific regression tracked upstream at duckdb/duckdb-wasm#2199.

### Where it bites in GeoLibre

- **Drag-and-drop / Add Data** of a `.parquet`/`.geoparquet` file (both web and desktop, since desktop also parses in the webview via DuckDB-WASM).
- **SQL workspace** remote reads, e.g. `read_parquet('https://…')`.

### Local finding beyond the upstream report

Upstream #2199 states extension load order is "not a factor." In our case it clearly is for the **local registered-buffer** path: a connection that runs `LOAD spatial` *before* its first `read_parquet` triggers the crash on `DESCRIBE`/`SELECT`. Reading the file once *before* `LOAD spatial` primes the connection and avoids it.

### Current workarounds in the code (what this issue tracks)

- `apps/geolibre-desktop/src/lib/duckdb-vector-loader.ts`
  - `parquetWarmUp()` (~L238) — pre-`LOAD` read `SELECT 1 FROM read_parquet(...) LIMIT 0`.
  - `ensureSpatialExtension()` (~L92) runs the warm-up before `LOAD spatial`.
  - `createDatabase()` (~L160) calls `db.open({})` so remote HTTP reads don't `stoi`-crash.
- `apps/geolibre-desktop/src/lib/sql-workspace.ts` — pre-spatial remote `read_parquet` warm-up + HTTPS-arg rewriting (~L124–L140, L739).

These are best-effort hacks. The warm-up can still fail for genuinely large or remote files, and it adds an extra read per first-load.

### Ask

Tracking issue so we can **remove the warm-up hacks and the `db.open({})` shim once upstream duckdb-wasm#2199 is fixed** and we bump the pin. Related: the native-desktop proposal would sidestep this entirely on desktop.

RohithPariki:
Fixes #961

## Root cause
The previous duckdb-wasm version (1.33.1-dev45.0) had a bug where loading the spatial extension before running `read_parquet` over HTTP would poison the instance, resulting in "stoi: no conversion" errors on subsequent queries. To work around this, several complex "warm-up" hacks and poison-recovery logic were added to `duckdb-vector-loader.ts` and `sql-workspace.ts`.

## Fix
Now that `duckdb-wasm` has been bumped to `1.33.1-dev57.0` (which includes the upstream fix for duckdb-wasm#2199), these workarounds are no longer necessary. 
Removed the `parquetWarmUp` and `beforeLoad` mechanisms from `ensureSpatialExtension`.
Removed the `isStoiConversionError` poison-recovery retry block and the `read_parquet` fallback warmup from `runSqlQuery`/`runSqlStatementOnce`.

## Verification
SQL Workspace queries using a remote Parquet file over HTTP will successfully load without any "stoi" errors, just as they would with the workarounds in place.

## Summary by CodeRabbit

* **Performance**
  * Reduced database startup overhead by removing unnecessary Parquet preloading and warm-up queries.
  * Simplified spatial extension initialization for faster workspace setup.

* **Bug Fixes**
  * Improved SQL workspace initialization by eliminating redundant remote data reads.
  * Removed automatic database rebuilding and query retries following a specific remote Parquet conversion error.


giswqs:
duckdb-wasm 1.33.1-dev57.0 does not resolve the issue. https://github.com/duckdb/duckdb-wasm/issues/2199 is still open. 

image

Issue #2475, opened 2026-09-18
[Tracking]: Mapbox renderer parity with the MapLibre renderer


giswqs:
A code review of the Mapbox renderer against the MapLibre renderer. Everything below comes from reading the code; nothing here was reproduced in a browser yet, so each item should be confirmed against an authenticated Mapbox map before it is fixed.

## Summary

- Every public `MapController` method except `init` is already on the `MapEngine` interface, so the engine interface is not where features go missing.
- About 90 app call sites do `mapControllerRef.current?.getMap()`. `getMap()` deliberately returns `null` on Mapbox, so each of them silently does nothing.
- `packages/map/src/MapboxCanvas.tsx` (271 lines) reimplements only a small part of `packages/map/src/MapCanvas.tsx` (2,505 lines).
- `compileMapboxLayer` (`packages/map/src/mapbox-layers.ts`) covers only the basic symbology subset of `layer-sync.ts`.

`docs/mapbox-renderer.md` already covers the intentional limits (MBTiles, Gaussian splats, the offline PMTiles basemap, COG terrain, blend modes, the Swipe raster provider, and similar). Everything below is undocumented.

## Tier 1: offered in the UI but broken or silent

- [x] **Story Map Present renders nothing.** The presenter returns null without `getMap()` and the entry is not gated, so the engine's `applyStoryChapterCamera` / `setStoryLayerOpacity` are never reached (`StoryMapPresenter.tsx:298,534`). The handout dialog and compose bar bail the same way. *(M)*
- [x] **Saved COG / raster layers are not restored** after a project reopen or renderer swap. The non-native-map branch restores rasters only for ArcGIS, yet `isMapboxPluginLayer` reports the layer as supported, leaving a Layers-panel row with nothing drawn and no error (`DesktopShell.tsx:1327-1337`). *(S)*
- [x] **Select features tools do nothing** (click, rectangle, polygon, freehand). Only `MapCanvas` listens for `FEATURE_SELECTION_EVENT`, and the menu is not gated (`LayerPanel.tsx:4100`, `MapCanvas.tsx:1393-1699`). *(M)*
- [x] **Field Collection, GPS Tracking, and the Georeferencer's map link** stop at a null `getMap()` (`FieldCollectionDialog.tsx`, `GpsTrackingDialog.tsx`, `GeoreferencerDialog.tsx:219-231`). *(M each)*
[...]

PR #2498, opened 2026-09-19
fix(map): restore MapLibre identify selection on popup dismissal (#2475)

giswqs:
## Summary

Part 2 of the identify-lifecycle unification (#2475): MapLibre's single-layer identify popup now uses the shared `map-identify-lifecycle` restore, matching Mapbox exactly. Per the maintainer's decision, closing an identify popup (X click or map click) now **restores the pre-identify selection** — both renderers finally share one behavior.

## What changed

- `MapCanvas.tsx` (single-layer hit path only):
  - New `showResolvedHitPopup()`: snapshots via the shared `createIdentifyPopupState` *before* the hit selection is applied; on genuine user dismissal the shared `restoreIdentifySelection` rolls back layer + feature (single and multi-select).
  - All other removal sites (follow-up hit, miss, mode cleanup, gesture start, project reload, abort/swap) now go through `removeIdentifyPopup()`, which detaches the close listener and removes the popup **without** restoring — programmatic removal never restores.
  - Selection effect now passes `isRestoringIdentifySelection()` instead of the hard-coded `false` to suppress camera fit during restore.
- `tests/map-identify-lifecycle.test.ts`: new test covering snapshot capture, user-dismiss restore, preserve-independently-changed, miss (no restore), abort/swap (no restore).

## Intentionally NOT touching (separate follow-ups)

Identify-All, WMS GetFeatureInfo, pixel/COG/NetCDF, and DuckDB identify paths — each has its own loading/abort mechanics.

## Validation

- `npm run test:frontend`: 8,836 passed, 0 failed
- `npm run build`: clean

## Authenticated MapLibre browser checks (needs a live map)

1. Select a feature; Identify another vector feature; close with X → original selection returns, no camera jump; close by clicking the map → same.
2. Identify A then click B: popup replaced cleanly; closing B restores pre-B selection.
3. While the popup is open, Ctrl/Shift-select in the Attribute Table → closing the popup keeps that selection (guard works).
4. Miss (no feature) → clears, no restore.
5. Turn Identify mode off with popup open → no restore.
6. Open another project with the popup open → no stale-project restore.
7. WMS / pixel / DuckDB / Identify-All flows unchanged.

## Note (known risk, flagged for review)

The camera-fit suppression during restore goes through the selection `useEffect`, which runs after the restore call's `finally` block may have reset the module flag — so that arg may be a no-op on MapLibre, unlike Mapbox (where the same call is synchronous inside a store subscription). Selection restore itself is covered by tests; the **camera-jump check (#1 above)** is the deciding observation; if it fires, the fix is to plumb the flag via the shared state rather than the module-level setter.

Closes part of opengeos/GeoLibre#2475 (MapLibre now matches Mapbox identify-dismiss behavior).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved identify popup behavior so dismissing a resolved popup restores the previous selection when appropriate.
  - Preserved independent user selections made while a popup is open.
  - Prevented programmatic popup replacement, cleanup, mode changes, or interrupted removal from unexpectedly restoring older selections.
  - Improved selection highlighting during restoration and prevented unnecessary map fitting.
  - Ensured selection restoration occurs only once and remains consistent across map views and update timing.


giswqs:
Not changing behavior here — this path now deliberately matches Mapbox, which is the goal of #2475's identify unification (maintainer's explicit decision: "MapLibre now matches Mapbox").

- **Miss** → clear the identify result: identical in `MapboxCanvas.tsx` (the miss branch calls `removeIdentifyPopup()` with no restore, then `selectFeature(null)`) and here (`clearIdentifyResult()`).
- **Replacement hit** → snapshots the previous identify state: identical in both (`showResolvedHitPopup` / the Mapbox `showIdentifyPopup` replacement both remove the prior popup **without restoring** and snapshot whatever selection is current, i.e. the previous identify result).

The alternative you're proposing (miss → restore pre-identify selection; replace → inherit the oldest pre-identify snapshot) would be a *third* behavior, requiring changes on **both** engines plus rethinking the "retire" semantics — worth a dedicated follow-up issue if desired, not a mid-stream behavior change in this PR. For now the two renderers are consistent, which closes the parity gap this item set out to fix.

giswqs:
Same position as in the thread on the prior revision: this is the intended, now engine-consistent behavior, not a defect. The miss branch in `MapboxCanvas.tsx` (around L409-413) does exactly `removeIdentifyPopup(); next.selectFeature(null);`, and its replacement path (around L417) drops the prior snapshot and re-snapshots the current selection — identical to this file. That parity is the explicit outcome of #2493 (MapLibre's "never undoes it" was the reported bug class being closed) and the maintainer's decision on #2498: "MapLibre now matches Mapbox." Making a miss restore the pre-identify selection would change the shared contract on both engines (Mapbox's miss currently clears) — a third behavior, best tracked as its own issue if desired; it is deliberately not part of this parity PR.

giswqs:
Understood — agreed a mounted-component or Playwright regression test is the right tool (the shared-module tests pin the marker's exactly-once semantics, but not the effect's consumption window). Adding one needs a decision on where it lives (there is no existing React-component test harness in this repo; E2E is Playwright against built apps), so I'd rather land this parity PR now and follow up with an identifying-popup e2e (identify → dismiss → assert no camera move + selection restored) than expand this PR's scope. I'll open that follow-up issue. For now the PR body documents the manual check list for the release notes.

giswqs:
Done in 0f5aafd2: shared the construction into a local createAndAddIdentifyPopup helper used by both paths; options now defined once.

giswqs:
@coderabbit ai re-review the latest commit (0f5aafd2): the DRY fix is in, and the remaining two points (miss/replace semantics; an e2e for camera fit) are deliberately deferred to a follow-up issue and documented in the PR. Please re-review and update your approval.

giswqs:
coderabbit ai review

giswqs:
Update: the two deferred points are now tracked — the E2E camera-fit/suppress regression test is issue #2500, and the miss/replace snapshot semantics are the shared-behavior discussion in this thread (parity with Mapbox, as decided) which I'd also like a dedicated issue if we want to change it. So for review purposes: the DRY item is fixed in 0f5aafd2, and the two design/test items are intentionally out of scope with follow-up issues. Please update your review when you get a chance.

PR #2502, opened 2026-09-19
test(map): lock Identify-restore fit-suppression (#2475)

giswqs:
test(map): adds a regression test pinning the Identify-restore fit-suppression contract between the shared marker and the selection effect for both Mapbox and MapLibre.

## Summary by CodeRabbit

- **Tests**
  - Added coverage for restoring map selections from popups without triggering an unnecessary map fit.
  - Verified that suppression applies only once and only to the matching selection.
  - Added safeguards for superseding, user-changed, and skipped-restore scenarios.
  - Covered selection-key serialization for single and multi-selection cases, including collision-resistant formats.
  - Tested both immediate and deferred map rendering paths to ensure consistent behavior.
  - Confirmed consistent handling across supported map rendering paths.


giswqs:
**Addressed** in ec9a8b6e — you're right. The previous main case only exercised the deferred read, so the marker assignment could be moved to *after* `next.selectFeatures(...)` and the test would still pass, breaking the synchronous Mapbox subscriber. Reworked:

- The main case now installs the exact shape MapboxCanvas uses — a `useAppStore.subscribe` observer whose body reads the marker for the current selection and calls `applySelectionHighlight` with the result, mirroring `MapboxCanvas.tsx:232-241`. Because zustand invokes the observer synchronously inside the `set()` behind `selectFeatures`, the marker must be written before that call for the observer to see it; the test asserts the fit-suppressed write happens and that the marker is already consumed before `restoreIdentifySelection` returns. A follow-up case covers the deferred (MapCanvas effect) path — consuming after the restore returns, matching `MapCanvas.tsx:1397-1399`.

Mutation-verified: temporarily moving the `pendingIdentifyRestoreKey = ...` assignment to *after* `next.selectFeatures(...)` makes the new synchronous case fail and the deferred one pass; restoring the file turns the whole green. I'll run the suite on CI to confirm the final state.

giswqs:
Both nits addressed in 8edf7f51 (unused imports removed; the dead recordingEngine() call removed from the superseding-selection test). Re-reviewing on the latest commit please.

giswqs:
Removed — createIdentifyPopupState and resolveHighlightIds are no longer imported in 8edf7f51.

giswqs:
Removed — the recordingEngine() call is gone from the superseding-selection test in 8edf7f51 (that case now only asserts marker reads, no engine interaction).

giswqs:
@coderabbitai review

giswqs:
@coderabbitai full review

giswqs:
Fixed in 3223897e — the deferred (MapCanvas effect) case now reads the restored field values back from the store (asserting selectedLayerId / selectedFeatureId / selectedFeatureIds), derives the fit key from that same store state, then consumes the marker for it. A hard-coded wrong restore write can no longer be masked by the marker.

giswqs:
Agreed — made it byte-for-byte: the subscriber now consumes the marker and applies the highlight on EVERY qualifying store write (including the intermediate null-key selectLayer write), exactly like MapboxCanvas, instead of skipping null-key writes. The sync case pins to the restored set specifically, so the intermediate write is no longer mistaken for the fit-suppressed one.

giswqs:
coderabbit ai review

giswqs:
Fixed in 14594733 — both cases now filter the matching calls and assert exactly one fit-suppressed restore (the sync case: exactly one [a, b] highlight; the deferred case: likewise), so a duplicated fit:false restore would now fail the test.

giswqs:
coderabbit ai review

giswqs:
coderabbit ai review

PR #2503, opened 2026-09-19
fit(map): honor metadata/source bounds and guard fitBounds on Mapbox

giswqs:
## What

Fix `fitBounds` and `fitLayer` on the Mapbox engine so they respect the layer's advertised bounds and no longer pass the map an impossible camera.

## Why (issue #2475)

Three concrete problems, all on the Mapbox engine while the MapLibre one
(`MapController.fitBounds` / `fitLayer`) already handles them:

1. **`maxZoom` was hard-coded to 14.** For an extent wider than a hemisphere
   a globe can show, mapbox-gl's globe fit answers by zooming *in*, leaving
   the data behind the horizon. The old call handed it 14 regardless. The
   MapLibre engine instead computes a flat-map ceiling with
   [`globeSafeMaxZoom`](../../packages/map/src/globe-fit-bounds.ts).
2. **No guard for non-finite or point-sized boxes.** `fitBounds([-10, -5,
   Infinity, 5])` was passed to the map unchanged, and a point-sized box
   (`w === e && s === n`) got a degenerate fit. The MapLibre engine drops
   non-finite coordinates and flies to the point for point-sized boxes.
3. **`fitLayer` only read live GeoJSON** via `getLayerBounds`, so a tile or
   raster layer that carries `metadata.bounds` (or only source-bounds) got no
   zoom-to-layer at all — "zoom to layer" silently did nothing for those
   layers. The MapLibre engine resolves
   `getLayerBounds ?? getLayerMetadataBounds ?? getLayerSourceBounds`.

## How

Mirror the MapLibre reference into `packages/map/src/mapbox-engine.ts`:

- `fitBounds`: guard non-finite coordinates; fly to a point at the floor zoom
  for point-sized boxes; replace `maxZoom: 14` with
  `globeSafeMaxZoom(bounds, getViewportSize(), FIT_BOUNDS_PADDING)`.
- `fitLayer`: extend bound resolution to `metadata.bounds` and then a
  source-bounds fallback from the engine's source plan.
- `layer-control-host.ts`: export `normalizeLayerBounds` so the engine can
  reuse it for the source-bounds fallback. The `getLayerMetadataBounds` helper
  is already exported and is now used here; no behavioral change to existing
  callers.

No other behavior is touched — selection highlighting, sync order, story
opacity, and the rest of the engine keep their existing tests.

## How to test

New `tests/mapbox-engine.test.ts` cases (all passing):

- **guards fitBounds: ignores non-finite and flies to a point-sized box**
  `fitBounds([-10, -5, Infinity, 5])` makes no camera move; a point-sized
  box produces a `flyTo` at the floor zoom (14), not a degenerate `fitBounds`.
- **stops the world-wide fit at a flat-map zoom instead of zooming in**
  With a laid-out 1000×600 viewport, `fitBounds([-180, -85, 180, 85])`
  records an option whose `maxZoom` is the dynamic ceiling (verified
  ≈ 0.027 via `globeSafeMaxZoom`), not the old hard-coded 14.
- **resolves fitLayer bounds from metadata and then the native source**
  A layer with no GeoJSON still fits when it advertises `metadata.bounds`,
  and (last resort) the engine falls through to the source-bounds the plan
  holds when there is no metadata either.

Full frontend suite: `8843 pass / 0 fail / 1 skipped` (the one skip is a
pre-existing, unrelated one). `oxfmt` and `eslint` clean.

## Notes

- This is the `#2475` row **"Mapbox: `fitBounds` / `fitLayer`: `maxZoom`
  hard-coded to 14, no fallback to metadata or source bounds (zoom-to-layer
  does nothing for tile layers), non-finite bounds not guarded."**
- The pre-existing `Cannot find module 'mapbox-gl/dist/mapbox-gl.css'`
  typecheck error is a CSS shim issue unrelated to this change; it
  also reproduces on `main` and is not a regression here.
- I did not run `pre-commit` (not installed locally); CI enforces it, and
  `oxfmt` + `eslint` + `tsc` are all clean for the changed files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved map viewport fitting for layers with metadata or source-provided boundaries.
  * Added fallback handling for boundaries defined by native map sources.
  * Prevented invalid, incomplete, or unavailable map boundaries from causing incorrect movement.
  * Improved handling of point-sized layers by centering the map at an appropriate zoom level.
  * Updated world and large-area views to use calculated zoom limits for more reliable framing.
  * Added safeguards for unavailable maps and zero-sized display areas.


giswqs:
@coderabbitai review

giswqs:
Agreed — removed. `getLayerBounds` already falls through to `source.bounds` → `metadata.bounds` with the same finite-number check, so the middle term was unreachable. The chain is now `getLayerBounds(layer) ?? this.getLayerSourceBounds(layer)`, and the source-plan fallback (the genuinely new path on Mapbox) is the last resort. I also reworded the test so it no longer implies `getLayerMetadataBounds` was the new piece.

giswqs:
Fixed in f3049fd8 — the assertion now requires `typeof options.maxZoom === "number" && options.maxZoom < 14`, so it fails if the ceiling is dropped entirely or falls back to the old constant. Verified: `globeSafeMaxZoom([-180,-85,180,85], 1000×600, 40)` returns ≈ 0.027, a number well below 14.

giswqs:
@coderabbitai review

giswqs:
Same root cause as the github-actions finding above — fixed in f3049fd8, now asserting `typeof options.maxZoom === "number" && options.maxZoom < 14`.

